PR #135, opened 2023-10-30
Remove Standard Surface approximation of coat darkening

jstone-lucasfilm:
This changelist removes the legacy Standard Surface approximation of the darkening and saturation of base colors under coatings.

In a future version of OpenPBR, our plan is to implement this coat darkening in a more subtle, physically based fashion, but the current approximation is too far from this ground truth to be useful.

### DIFF
--- a/reference/open_pbr_surface.mtlx
+++ b/reference/open_pbr_surface.mtlx
@@ -168,39 +168,14 @@
       <input name="in2" type="vector3" interfacename="geometry_tangent" />
     </ifgreater>
 
-    <!-- Colors influenced by coat ("coat gamma") -->
-    <clamp name="coat_clamped" type="float">
-      <input name="in" type="float" interfacename="coat_weight" />
-    </clamp>
-    <multiply name="coat_gamma_multiply" type="float">
-      <input name="in1" type="float" nodename="coat_clamped" />
-      <input name="in2" type="float" value="1.0" />
-    </multiply>
-    <add name="coat_gamma" type="float">
-      <input name="in1" type="float" nodename="coat_gamma_multiply" />
-      <input name="in2" type="float" value="1.0" />
-    </add>
-    <max name="base_color_nonnegative" type="color3">
-      <input name="in1" type="color3" interfacename="base_color" />
-      <input name="in2" type="float" value="0.0" />
-    </max>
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Here's a simple OpenPBR Carpaint material rendered using this latest reference implementation of the model:

OpenPBR_Carpaint_Latest

portsmouth:
How about we instead just make the `coat_affect_color` value smaller than 1, say 0.5?

jstone-lucasfilm:
There were some great notes from @peterkutz in the related Slack thread that seem relevant here:

> I have done these simulations in the past, although I haven't had a chance to do new tests in light of our recent discussions. I do know from the past that the magnitude of the effect depends on multiple factors, such as the roughness, IOR, color, viewing angle, etc. For example:
> - in some cases the base color appears darker and more saturated (e.g., if the coat has a higher IOR than the base surface and the base color has a medium brightness)
> - in some cases the base color is unaffected (e.g., if the base surface is black or white, or if the coat has the same IOR as the base surface)
> - in some cases the base color actually appears brighter (e.g., if the coat has a lower roughness than the base surface resulting in less multiple scattering and less reflection from the coat, or if the coat has a lower IOR than the base surface resulting in less overall specular reflection from the combination of the coat and the base surface (in this case, the coat functions as a [graded-index anti-reflective coating](https://en.wikipedia.org/wiki/Anti-reflective_coating)))

Given our expectation that an accurate implementation will have a highly variable effect, sometimes darkening and sometimes lightening the underlying base color, perhaps our best short-term approach is to "do no harm" and remove the uniform squaring effect that was used in Standard Surface?

I'm definitely open to other ideas here, but the current approximation seems too far from the ground truth to be a "rough draft" of our future implementation, and I'm concerned about this unintuitive behavior leaving artists with a negative first impression of OpenPBR.

AdrienHerubel (approved):
The consensus when discussing this PR with @portsmouth and @virtualzavie was that the current approximation is too simple and should be replaced by a better view dependant and possibly roughness dependant version which being discussed in https://github.com/AcademySoftwareFoundation/OpenPBR/issues/136. 

In the meantime instead of using the Standard Surface approximation without a way for artists to disable it, we should remove it from the implementation.

Issue #148, opened 2023-12-05
MaterialX reference implementation errors


JenusL:
While trying out OpenPBR I noticed that there's some errors in the MaterialX reference implementation.
- `transmission_dispersion` should be removed.
- `transmission_dispersion_abbe_number` is missing.
- `transmission_dispersion_scale` is missing.
- `transmission_scatter_anisotropy` has `uimin` set to 0, this should be -1 so that it can allow for back scattering.

portsmouth:
Fixed this up in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/151.

jstone-lucasfilm:
Thanks for posting this issue, @JenusL, and thanks to @portsmouth for the fix!

PR #151, opened 2024-01-11
Fix transmission implementation and improve specification

portsmouth:
Also improve wording/formatting in the spec Translucent base section.

This addresses the points raised in https://github.com/AcademySoftwareFoundation/OpenPBR/issues/148.

### DIFF
--- a/reference/open_pbr_surface.mtlx
+++ b/reference/open_pbr_surface.mtlx
@@ -39,7 +39,7 @@
            doc="The amount of directional bias, or anisotropy, of the volumetric scattering in the transparent base." />
     <input name="transmission_dispersion_scale" type="float" value="0.0" uimin="0.0" uimax="1.0" uiname="Transmission Dispersion Scale" uifolder="Transmission" uiadvanced="true"
            doc="Linearly scales the amount of dispersion." />
-    <input name="transmission_dispersion_abbe_number " type="float" value="0.0" uimin="0.0" uisoftmin="9.0" uisoftmax="91.0" uiname="Transmission Dispersion Abbe Number" uifolder="Transmission" uiadvanced="true"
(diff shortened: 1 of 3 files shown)


reference/open_pbr_surface.mtlx
```diff
@@ -35,10 +35,12 @@
...
+    <input name="transmission_scatter_anisotropy" type="float" value="0.0" uimin="-1.0" uimax="1.0" uiname="Transmission Anisotropy" uifolder="Transmission" uiadvanced="true"
            doc="The amount of directional bias, or anisotropy, of the volumetric scattering in the transparent base." />
-    <input name="transmission_dispersion" type="float" value="0.0" uimin="0.0" uisoftmax="100.0" uiname="Transmission Dispersion" uifolder="Transmission" uiadvanced="true"
-           doc="Dispersion amount, describing how much the dielectric index of refraction varies across wavelengths." />
+    <input name="transmission_dispersion_scale" type="float" value="0.0" uimin="0.0" uimax="1.0" uiname="Transmission Dispersion Scale" uifolder="Transmission" uiadvanced="true"
+           doc="Linearly scales the amount of dispersion." />
+    <input name="transmission_dispersion_abbe_number " type="float" value="0.0" uimin="0.0" uisoftmin="9.0" uisoftmax="91.0" uiname="Transmission Dispersion Abbe Number" uifolder="Transmission" uiadvanced="true"
```
jstone-lucasfilm, line 42: Just a minor typo, there's a space in the new input name that generates warnings when the shading model is included in MaterialX environments.

jstone-lucasfilm (commented):
This looks like a good change, @portsmouth, though we should address the warning noted below.

jstone-lucasfilm (approved):
This looks good to me, @portsmouth, and most of the OpenPBR example materials work correctly with this change, though we have one example material that needs an update to render correctly:
https://github.com/AcademySoftwareFoundation/OpenPBR/blob/main/examples/open_pbr_surface_default.mtlx

I'll go ahead and approve this change, but we should make sure to fix and verify the example materials before marking an 0.3 release.

Issue #141, opened 2023-11-14
Glossy-diffuse slab description as a gloss layer is possibly confusing


portsmouth:
We currently write in the spec:

>The glossy-diffuse slab represents a dielectric with rough GGX microfacet surface BSDF, embedding a semi-infinite bulk of extremely dense scattering material.

This is the model we want to use physically, i.e. the slab is like the infinitely dense limit of the subsurface.

But then we say:
>We choose to model this concretely as a layer of dielectric “gloss” on top of an opaque slab with a diffuse BRDF:

The reason we opted to describe glossy-diffuse as `layer(diffuse, gloss)` is then at least it's clear what the base color/roughness mean (i.e. the color and roughness of the Oren-Nayar base).

If you want to think of it instead as a dense subsurface inside dielectric, it's really not clear what base color/roughness should mean for the properties of the subsurface (e.g. there is no more Oren-Nayar model going on, so what is the roughness doing -- something like "rough volume", but that's not standard). Also if we did that with base color being the volume albedo say, technically the scattering would cause a color shift, requiring remapping (like the actual subsurface, but in some infinite density limit).

With the existing description saturation should occur for this layer too, due to the bounces within it. (And the coat would just add further saturation). That's not completely unrealistic, since even in the volumetric model some saturation will happen due to TIR and multiple scattering.

So I actually don't know what the best way to describe it is. Possibly the existing description is the best option given the currently available models. We might want to say though that we assume some remapping is done so that the supplied base color appearance is produced in some sense. How to define that so it is not incoherent I'm not sure though.

peterkutz:
> We might want to say though that we assume some remapping is done so that the supplied base color appearance is produced in some sense.

This type of approach seems reasonable to me. It seems like "diffuse reflection" can be caused by a variety of different underlying mechanisms (e.g., microfacets, volumetric scattering, spongy internal structure, etc.) and from an artist's perspective it doesn't seem that important which one is happening. The important thing seems to be that there's some dielectric substance that exhibits specular reflection from its outer surface and diffuse reflection from below that. The overall vdiffuse reflection has some user-specified color, even though the underlying scattering events that lead to the diffuse distribution of light might have different colors. At a high-level, I feel like it's fine for the "roughness" to define the overall shape of the diffuse lobe without enforcing some low-level physical structure, but I see the problem of making the specification rigorous and clear.

[...]

Issue #136, opened 2023-10-31
Better approximation for darkening layers under coat


AdrienHerubel:
Due to the possibility of total internal reflection at the coat/external medium boundary, layers under the coat should be darker. See https://graphics.cs.yale.edu/sites/default/files/wet.pdf for a detailed model describing the various effect of having a water layer on top of materials. I think that interpenetration of layers is out of scope for OpenPBR surface, but the total internal reflection should be modelled.

In the OpenPBR specification we describe the phenomenon without providing implementation details :
`In the full light transport this observed color is further darkened and saturated due to multiple internal reflections from the inside of the coat, including a considerable amount of total internal reflection, which causes light to strike the underlying material multiple times and undergo more absorption. Also the observed tint color should vary away from coat_color as the incidence angle changes, due to the change in path length in the medium. The presence of a rough coat will increase the apparent roughness of the BSDF lobes of the underlying base. We generally assume that in the ground truth appearance, all these effects are accounted for.`

In Standard Surface there is a dedicated parameter to simulate the darkening of layers under coat using the following formula :
`base_color = pow(base_color, 1.0 + (coat * coat_affect_color)) subsurface_color = pow(subsurface_color, 1.0 + (coat * coat_affect_color)) 
`
This approximation is simple and disabled by default in Standard Surface. 

In OpenPBR we should evaluate it against ground truth, possibly find a better approximation, and decide whether we want to make the effect user controllable.

portsmouth:
For the moment, I think it's fine to use the crude Standard Surface model, since according to the spec an implementation can approximate the effect however it wants (including just ignoring it..).

I'd like to leave the effect on since the spirit of the spec is to do this physically correctly, in which case there should be a darkening/saturation effect. Even with the rough approximation, it will be good to get feedback about this and because (as Indy argued) it probably is generally more useful for the effect to be on than off. 

I have a slight preference for using a lower hard-coded value than 1 though (e.g. 0.5), since the effect is quite strong right now, and that may not be realistic.

As noted in the OpenPBR thread, I think we should for a later release derive a slightly better approximation to get more insight into how big this effect actually is. For example if you restrict to normal incidence, I think the full multi-bounce calculation is pretty tractable.

Whether we put in some control later to modulate the effect, is tricky since we would have to give some concrete physical explanation of it.  But once we have a better understanding of a formula for the effect, we could implement it for example like:

[...]

PR #162, opened 2024-02-16
Add coat darkening and saturation effects

portsmouth:
This PR adds a detailed discussion to the spec of the coat darkening effect (incorporating the ideas discussed in https://github.com/AcademySoftwareFoundation/OpenPBR/issues/136 and https://github.com/AcademySoftwareFoundation/OpenPBR/issues/141), with:
  - introduction of a new `coat_darkening` parameter (defaulting to 1, the physically correct case).
  - a physical definition of how this parameter operates to reduce the darkening (via boosting of the base albedo)
  - a suggested practical implementation of the darkening factor for the base lobe, taking into account the IOR of the coat, and also the roughness of the base (based on the derivations and discussions linked elsewhere).
  - A suggested implementation of the darkening of colored coats at grazing angles due to longer path length in the absorbing medium.

I also include a placeholder for a discussion of how the coat roughness should affect the underlying lobes. (Probably to be done in a separate PR, and derivation of a reasonable physical approximation for this is WIP).

To make it easier to read, I broke the Coat section into sub-sections.

image

image

image

image


portsmouth:
One aspect of the proposed formulas that I worry about is the effect of the base roughness on the darkening, i.e. Equation 61. This is designed so that a smooth base produces less darkening,  basically blending between known formulas for the cases of a completely smooth and completely diffuse base.

I assumed that if the base is dielectric, the roughness of the reflection from it is determined by the `specular_roughness`. However if the base has a scattering medium (i.e. as it does in the glossy-diffuse case, or if there is a transmission or subsurface volume) then the reflection is effectively roughened due to this scattering, apart from the initial Fresnel reflection.  The proposed formula ignores this effect, so probably underestimates the darkening effect of the coat on a glossy-diffuse base or subsurface. We would need to do some proper investigation with Monte Carlo simulations to know.

Perhaps it's fine to still give these formulas merely as a convenient, reasonable approximation.

index.html
```diff
@@ -857,6 +852,91 @@
...
+While this defines the behavior at a physical level, in practice implementations will need to develop some specific approximation for the coat darkening effect.
+While this can be implementation dependent, we suggest here a reasonably simple, efficient scheme which captures the essential behavior.
+
+It can be shown that in the case of a Lambertian base with (constant) albedo $E_b$ and a perfectly smooth dielectric clear-coat -- the so-called _interfaced Lambertian_ model ([#Elias2001], [#d'Eon2021]) -- that the exact directional albedo of the coated base has the albedo-scaling-like form (where $\eta_c$ is the ratio of the coat IOR to the ambient IOR, and $F$ is the corresponding coat Fresnel factor):
+\begin{equation} \label{interfaced_lambertian_albedo}
+E_c(\omega_o) = F(\omega_o, \eta_c) + E_b \Bigl(1 - F(\omega_o, \eta_c)\Bigr) \Delta(E_b, \eta_c)
+\end{equation}
+where $\Delta(E_b, \eta_c)$ is a darkening factor given by
+\begin{equation} \label{general_darkening_formula}
+\Delta(E_b, \eta_c) = \frac{1 - K}{1 - E_b K} \ .
```
peterkutz, line 882: I haven't looked at the sources yet but I'm assuming the &\Delta& terminology and definition comes from there.

index.html
```diff
@@ -857,6 +852,91 @@
...
+\begin{equation} \label{internal_diffuse_reflection_coefficient_for_smooth_base}
+K_s = F(\omega_o, \eta_c) \ ,
+\end{equation}
+which becomes exact in the limit of a perfectly smooth metallic or dielectric base. There is significantly less darkening in the case of a smooth base, as the specular reflection from the base is less likely to generate total internal reflection at the coat boundary.
+
+We thus recommend in the general case that the darkening factor of equation [general_darkening_formula] be applied, but taking
+\begin{equation} \label{internal_diffuse_reflection_coefficient_for_general_base}
+K = \mathrm{lerp}(K_s, K_r, r_b)
+\end{equation}
+where $r_b$ is an estimate of the base roughness, a blend between metal $r_m$ and dielectric $r_d$ roughness estimates according to the **`base_metalness`** $\mathtt{M}$:
```
peterkutz, line 901: > a blend between metal $r_m$ and dielectric $r_d$ roughness estimates

Readability might be improved by swapping the order of $r_m$ and $r_d$ to match the following formula:

> $r_b = \mathrm{lerp}(r_d, r_m, \mathtt{M})$

index.html
```diff
@@ -857,6 +852,91 @@
...
+\begin{equation} \label{internal_diffuse_reflection_coefficient_for_rough_base}
+K_r = 1 - \Bigl(1 - E_F(\eta_c)\Bigr)/\eta_c^2 \ .
+\end{equation}
+The albedo $E_F(\eta_c)$ is the _hemispherical_ (or _average_) albedo of the coat Fresnel factor [^avg_fresnel]. If the base has specular reflection (due to a smooth metallic or dielectric interface) then equation [general_darkening_formula] still applies, but the appropriate value of the internal diffuse reflection coefficient $K$ should be closer to:
+\begin{equation} \label{internal_diffuse_reflection_coefficient_for_smooth_base}
+K_s = F(\omega_o, \eta_c) \ ,
+\end{equation}
+which becomes exact in the limit of a perfectly smooth metallic or dielectric base. There is significantly less darkening in the case of a smooth base, as the specular reflection from the base is less likely to generate total internal reflection at the coat boundary.
```
peterkutz, line 895: Nice descriptions.

index.html
```diff
@@ -857,6 +852,91 @@
...
+
+We thus recommend in the general case that the darkening factor of equation [general_darkening_formula] be applied, but taking
+\begin{equation} \label{internal_diffuse_reflection_coefficient_for_general_base}
+K = \mathrm{lerp}(K_s, K_r, r_b)
+\end{equation}
+where $r_b$ is an estimate of the base roughness, a blend between metal $r_m$ and dielectric $r_d$ roughness estimates according to the **`base_metalness`** $\mathtt{M}$:
+\begin{equation} \label{base_roughness_estimate}
+r_b = \mathrm{lerp}(r_d, r_m, \mathtt{M}) \ .
+\end{equation}
+The base dielectric roughness $r_d$ can reasonably be estimated as a mix between the high roughness of an assumed underlying base and the microfacet dielectric roughness $r$ = **`specular_roughness`**, according to the base dielectric Fresnel factor modulated via **`specular_weight`** $\xi_s$ (see equation [unmodulated_fresnel] and equation [modulated_ior]):
```
peterkutz, line 905: I'm assuming this Fresnel factor uses the relative IOR of the base and the coat (and not just the base IOR). Is that clear here? Or would that make things more complex (e.g., due to the possibility or TIR arising here)?

index.html
```diff
@@ -857,6 +852,91 @@
...
+\begin{equation} \label{dielectric_roughness_estimate}
+r_d = \mathrm{lerp}(1, r, \xi_s F_s)
```
peterkutz, line 907: Is this $F_s$ angle-dependent? If so, is it based on the original incident direction or the direction after being refracted into the coat?

index.html
```diff
@@ -857,6 +852,91 @@
...
+
+Here $K \in [0,1]$ is the _internal diffuse reflection coefficient_, which corresponds to the fraction of the energy leaving the base which returns
+to the base due to internal reflection from the coat. Higher values of $K$ lead to more internal reflections, thus more darkening and saturation. In the case of a Lambertian base (which should be a reasonable approximation to the rough metal, dielectric, or diffuse cases), $K = K_r$ where, with relative coat IOR $\eta_c$:
+\begin{equation} \label{internal_diffuse_reflection_coefficient_for_rough_base}
+K_r = 1 - \Bigl(1 - E_F(\eta_c)\Bigr)/\eta_c^2 \ .
+\end{equation}
+The albedo $E_F(\eta_c)$ is the _hemispherical_ (or _average_) albedo of the coat Fresnel factor [^avg_fresnel]. If the base has specular reflection (due to a smooth metallic or dielectric interface) then equation [general_darkening_formula] still applies, but the appropriate value of the internal diffuse reflection coefficient $K$ should be closer to:
+\begin{equation} \label{internal_diffuse_reflection_coefficient_for_smooth_base}
+K_s = F(\omega_o, \eta_c) \ ,
```
peterkutz, line 893: I'm assuming this $F$ could be dielectric or metallic depending on the value of $base_metallic$?

index.html
```diff
@@ -857,6 +852,91 @@
...
+\end{equation}
+which becomes exact in the limit of a perfectly smooth metallic or dielectric base. There is significantly less darkening in the case of a smooth base, as the specular reflection from the base is less likely to generate total internal reflection at the coat boundary.
+
+We thus recommend in the general case that the darkening factor of equation [general_darkening_formula] be applied, but taking
+\begin{equation} \label{internal_diffuse_reflection_coefficient_for_general_base}
+K = \mathrm{lerp}(K_s, K_r, r_b)
+\end{equation}
+where $r_b$ is an estimate of the base roughness, a blend between metal $r_m$ and dielectric $r_d$ roughness estimates according to the **`base_metalness`** $\mathtt{M}$:
+\begin{equation} \label{base_roughness_estimate}
+r_b = \mathrm{lerp}(r_d, r_m, \mathtt{M}) \ .
```
peterkutz, line 903: I would have to think about it more, but on first read something feels potentially slightly redundant here.

This formula is computing the overall roughness $r_b$ by blending two roughnesses based on $M$. Then the overall $K$ is computed by blending together $K_s$ and $K_r$ based on $r_b$. But it seems like $K_s$ must have also taken $M$ into account in order to blend between the dielectric and metallic $F$ in $K_s = F(\omega_o, \eta_c)$.

Intuitively it feels like $M$ should only need to appear once in the formulation, for blending the effective roughness of the other components with the effective roughness of the metallic component. Although then maybe the `specular_roughness` $r$ would need to appear twice instead.

In any case, the high-level approach of computing an effective roughness value makes a lot of sense. (We do something very similar to this in Eclair.)

index.html
```diff
@@ -1381,6 +1463,19 @@
 
 [^normalization]:  Omitting normalization factors.
 
+[^porosity]: In reality, coats can also darken the underlying surface due to a different mechanism where the coat modifies the Fresnel factor of the base due to the coat material filling in air gaps between granules or threads of a porous base material, which reduces the relative IORs at the internal interfaces. This occurs e.g. on adding water to sand or fabric, or adding a penetrating wood finish. We assume here that this effect explicitly does _not_ occur, at present, since we do not have enough knowledge about the properties of the underlying substance to model it. We can only safely assume that the first mechanism of darkening, i.e. internal reflections, occurs.
+
+[^avg_fresnel]: The hemispherical (or average) albedo of the Fresnel factor is defined as $E_F(\eta) \equiv 2 \int_0^1 F(\mu, \eta)\,\mu\,\mathrm{d}\mu$. This can be tabulated, or one can use the convenient analytical approximation [#d'Eon2021] (accurate to within 0.2% in the limited range $\eta \in [1,3]$):
+\begin{equation}
+  E_F(\eta) \approx \ln \biggl( \frac{10893\eta - 1438.2}{-774.4\eta^2 + 10212\eta + 1} \biggr) \ .
```
peterkutz, line 1470: Useful to have this reproduced here.

peterkutz:
> I assumed that if the base is dielectric, the roughness of the reflection from it is determined by the specular_roughness. However if the base has a scattering medium (i.e. as it does in the glossy-diffuse case, or if there is a transmission or subsurface volume) then the reflection is effectively roughened due to this scattering, apart from the initial Fresnel reflection. The proposed formula ignores this effect, so probably underestimates the darkening effect of the coat on a glossy-diffuse base or subsurface.

Is this not addressed by the formula below?

> $r_d = \mathrm{lerp}(1, r, \xi_s F_s)$

I thought this was considering the diffuse component to have a roughness of $1$.

peterkutz:
This is looking really good to me at a high level but I would need to do another pass to verify some of the formulas by looking more closely and by looking at the referenced equations and citations. A lot of my comments are questions and clarifications that I could potentially answer myself based on a more careful reading, however I figured I would post them anyway in case they raise any important points and in case the text can be clarified to prevent other readers from initially having the same questions.

index.html
peterkutz: I'm also assuming this formula is the closed form of a geometric series. Is that the case?

portsmouth: >I haven't looked at the sources yet but I'm assuming the $\Delta$  terminology and definition comes from there.

No actually, I just used $\Delta$ for the "darkening factor" since it's a Greek D..   Hopefully it seems reasonable anyway. For us, this darkening factor is a relevant quantity, but not really in the references.

>I'm also assuming this formula is the closed form of a geometric series. Is that the case?

Yes I think it can be derived via a geometric series (considering the "ladder" of bounces within the layer) hence the form of it, though in the Elias reference it's derived using some radiative transfer arguments so the series is not made explicit.

Here is my equations (57), (58) in the form shown in the Elias paper (which is essentially the directional albedo, excluding the Fresnel reflection):

image

where for them:
  - $\rho$ is the base albedo (they assume Lambertian)
  - $r_i$ is the "internal diffuse reflection coefficient", that I call $K$.
  - $r_e$ ("external diffuse reflection coefficient") is just the average Fresnel factor.
  - $T$ is a transmission coefficient, so reduces to $1 - F$ with some Fresnel factor $F$.

Hopefully the correspondence is not too hard to make. (My notation is consistent with the more detailed write-up I did).

index.html
portsmouth: It should be clarified that by Fresnel factor here I meant specifically the F0 from equation 25.

I'm not sure it's clear, but the Fresnel factor (of equation 25) is supposed to take into account the presence of the coat (so e.g. goes to zero if the coat and base dielectric have equal IOR). 

Yes TIR could occur, if the coat IOR > base IOR. In which case this Fresnel factor goes to 1, so we would take the base roughness to be `specular_roughness` in computing the darkening factor.

The idea of using the Fresnel factor in the lerp (for the dielectric base roughness estimate) was just that the dielectric specular reflection will be superimposed on some diffuse reflection, and the F0 seems a reasonable proxy for how much we should weight the dielectric roughness in the blend. Obviously, this is just a rough heuristic, attempting to capture the essence of the effect (that a smooth base darkens less) in as simple as possible a way.

index.html
portsmouth: No it's supposed to be the F0 of equation 25. That isn't clear in the text, so will fix. That was just a choice though, since as noted
>the F0 seems a reasonable proxy for how much we should weight the dielectric roughness in the blend.

Someone else could have a better idea for how this should work.

index.html
portsmouth: See the point above about the Fresnel factor in $K_s = F(\omega_o, \eta_c)$ referring to the coat dielectric, not the base (dielectric or metal).

index.html
portsmouth: There is also a more accurate approximation that works over a larger range of IORs, but it's much more lengthy and involves logs..

portsmouth:
> Is this not addressed by the formula below?
> 
> > rd=lerp(1,r,ξsFs)
> 
> I thought this was considering the diffuse component to have an effective roughness of 1.

I stated it incorrectly, I meant that if scattering medium is actually **absent**  -- e.g. coat on clear glass (or glass with a volume with huge MFP),  then it seems wrong to assume there is a diffuse component with a roughness of 1, that we should blend in according to the F0 of the glass.

Fixing that would involve the MFP somehow, but it would get a bit too elaborate probably.

portsmouth:
>This is looking really good to me at a high level but I would need to do another pass to verify some of the formulas by looking more closely and by looking at the referenced equations and citations. A lot of my comments are questions and clarifications that I could potentially answer myself based on a more careful reading, however I figured I would post them anyway in case they raise any important points and in case the text can be clarified to prevent other readers from initially having the same questions

Thanks, I appreciate the careful reading and discussion! :smiley:

portsmouth:
Not sure it's helpful, but my [implementation ](https://github.com/portsmouth/OpenPBR-viewer/blob/b7545a91dfcc854c3a711225ae0548621d5c3529/glsl/openpbr_surface.glsl#L91C1-L110C132)of this in the GLSL viewer might be a useful point of reference. It looks like this, where the darkening is applied to the weight of the "base substrate" lobe (which consists of the mix of metal and dielectric base lobes):
```glsl
    // Compute base darkening factor due to presence of coat:
    vec3 base_darkening = vec3(1.0);
#ifdef COAT_ENABLED
    if (coated && coat_darkening > 0.0)
    {
        float Kr = 1.0 - (1.0 - average_dielectric_fresnel(coat_ior))/sqr(coat_ior);
        float Ks = FresnelDielectricReflectance(abs(winputL.z), coat_ior);
        float Fr = clamp(specular_weight * fresnel_refl_normal_incidence(), 0.0, 1.0);
        float rd = mix(1.0, specular_roughness, Fr); // estimate of roughness of dielectric base
        float rm = specular_roughness;               // estimate of roughness of metallic base
        float rb = mix(rd, rm, M);  // thus estimated roughness of entire base
        float K = mix(Ks, Kr, rb);  // thus estimated internal diffuse reflection coeff.
        vec3 E_dielectric_base = mix(mix(albedos.m[ID_DIFF_BRDF], albedos.m[ID_SSSC_BTDF], S),
                                         albedos.m[ID_SPEC_BTDF], T);     // dielectric base albedo
        vec3 E_base = mix(E_dielectric_base, albedos.m[ID_META_BRDF], M); // entire base albedo
        vec3 Delta = (1.0 - K) / vec3(1.0 - E_base*K);                    // full darkening factor
        base_darkening = mix(vec3(1.0), Delta, C * coat_darkening);       // modulated darkening factor
    }
#endif // COAT_ENABLED
    vec3 w_base_substrate = w_coated_base * mix(vec3(1.0), base_darkening * coat_color * (vec3(1.0) - albedos.m[ID_COAT_BRDF]), C);
```

portsmouth:
I'm not sure it's practical to implement this in the MaterialX graph. It would be very time consuming to construct all the math in XML in without errors. 

To implement this in the MaterialX graph would require making various simplifications, as:
   - the view direction is not known in the graph (needed for $K_s$), so $K_s$ would have to be evaluated at normal incidence instead say
   - the dielectric Fresnel average integral would have to be coded up manually in the graph, which is a bit painful
   - the BSDF albedos are not possible to evaluate, so would need to make some approximation 
  
I can try to do something minimal, but probably the relevant functions should instead be exposed as built-in utility nodes.

jstone-lucasfilm:
@portsmouth Would it be practical to implement this logic in vanilla OSL, using only the set of closures and functions that OSL natively provides?  If so, then I could take the step of translating this back to a MaterialX graph that generates the vanilla OSL code.

portsmouth:
> @portsmouth Would it be practical to implement this logic in vanilla OSL, using only the set of closures and functions that OSL natively provides? If so, then I could take the step of translating this back to a MaterialX graph that generates the vanilla OSL code.

It would certainly be a lot easier than trying to write code in XML, yes. 

If the native functions are the ones listed [here](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/main/src/doc/osl-languagespec.pdf), then it doesn't have them built in (i.e. the formula for dielectric Fresnel factor and its hemispherical albedo). It's also not possible I take it to compute the directional (or hemispherical) albedo of a BSDF closure in OSL (or the MaterialX graph), so those albedos will I assume just have to be be approximated ignoring the view direction.

jstone-lucasfilm:
@portsmouth Yes, I believe that's the very latest specification for OSL, so that's the set of closures and functions that we would need to use here.  Would you like to put together a proposal for a pure OSL definition of the behavior you'd like to see, and I can do the work of translating this to the equivalent MaterialX graph?

portsmouth:
@jstone-lucasfilm 

Here's a first pass at the OSL code, expressed in a self-contained form as a function for the coat darkening factor (which will be multiplied into the base closure combination in the graph), taking as input the relevant subset of the OpenPBR parameters:

```C
// Hemispherical (average) albedo of dielectric Fresnel factor
float E_F(float eta)
{
    return log((10893.0*eta - 1438.2)/(-774.4*sqr(eta) + 10212.0*eta + 1.0)); // (NB, this is natural log)
}

float average_dielectric_fresnel(float eta)
{
    if      (eta > 1.0) return E_F(eta);
    else if (eta < 1.0) return 1.0 - sqr(eta)*(1.0 - E_F(1.0/eta));
    else return 0.0;
}

float dielectric_F0(float eta)
{
   return sqr((eta - 1.0)/(eta + 1.0));
}

color coat_darkening_factor(color base_color, 
                            float base_metalness,
                            float specular_weight,
                            float specular_roughness,
                            float specular_ior,
                            float transmission_weight,
                            color subsurface_color,
                            float subsurface_weight,
                            float coat_ior,
                            float coat_weight,
                            float coat_darkening)
{ 
   // Compute estimate of roughness of base
   const float ambient_ior = 1.0;                                                // (Assume here no nested dielectrics, for simplicity)
   float coat_ior_average = mix(ambient_ior, coat_ior, coat_weight);             // (as described in Coat section of spec)
   float eta_s = specular_ior / coat_ior_average;                                // (as described in Coat section of spec)
   float Fr = clamp(specular_weight * dielectric_F0(eta_s), 0.0, 1.0);           // dielectric specular reflection F0, modulated by specular weight
   float rd = mix(1.0, specular_roughness, Fr);                                  // estimate of roughness of dielectric base
   float rm = specular_roughness;                                                // estimate of roughness of metallic base
   float rb = mix(rd, rm, base_metalness);                                       // thus estimated roughness of entire base
   // Compute "internal diffuse reflection coefficient", taking into account estimated base roughness
   float coat_eta = coat_ior / ambient_ior;
   float Kr = 1.0 - (1.0 - average_dielectric_fresnel(coat_eta))/sqr(coat_eta);  // internal diffuse reflection coefficient, rough case
   float Ks = dielectric_F0(coat_eta);                                           // internal diffuse reflection coefficient, smooth case
                                                                                 // (approx., since we don't have access to wi in the MaterialX graph)
   float K = mix(Ks, Kr, rb);                                                    // thus estimated internal diffuse reflection coeff.
   // Compute estimated total base albedo
   color diffuse_albedo      = base_color;                                       // approx. diffuse slab albedo
   color subsurface_albedo   = subsurface_color;                                 // approx. subsurface slab albedo
   color transmission_albedo = Fr;                                               // approx. translucent slab albedo (ignore volume or tint, for simplicity)
   color metal_albedo        = base_color;                                       // approx. metal slab albedo
   color albedo_opaque_dielectric = mix(diffuse_albedo,           subsurface_albedo,   subsurface_weight);
   color albedo_dielectric        = mix(albedo_opaque_dielectric, transmission_albedo, transmission_weight);
   color albedo_base              = mix(albedo_dielectric,        metal_albedo,        base_metalness);
   // Final darkening factor
   color Delta = color(1.0 - K) / (color(1.0) - albedo_base*K);
   return mix(color(1.0), Delta, coat_weight * coat_darkening);                  // modulate according to coat weight
}
```

jstone-lucasfilm:
Thanks @portsmouth, and that looks straightforward to implement in MaterialX as a collection of new subgraphs, which we can define in the same MaterialX document as the reference graph for OpenPBR.  This will definitely add a significant amount of new code to generated shaders (e.g. GLSL, OSL, MDL), but it seems worthwhile to add this new logic and then evaluate the visual benefits versus the performance impact.

jstone-lucasfilm:
Since this is dependent on PR #157, let's finalize and merge that one first, and then we can start on the MaterialX graph for this proposal.

jstone-lucasfilm:
I've now merged #157, and the next step would be to resolve conflicts between these two pull requests, after which we can start on the MaterialX-side updates for this change.

portsmouth:
Conflicts resolved.

portsmouth:
@jstone-lucasfilm are you still intending to generate the MaterialX graph from the OSL code above?

jstone-lucasfilm:
@portsmouth I was planning to take this step in preparation for a v1.0 release in late June, which would give us sufficient time for validation and adjustments.  But if we need to lock down the codebase in just one week (as suggested by your colleagues in our last meeting), then this would become a significant last-minute change in behavior, which seems less practical to accomplish.

portsmouth:
We could presumably “lock down” the spec (as far as parameters and ranges) including the darkening, without the MaterialX implementation of the suggested math. If need be, we could put back the old standard surface approximation for the darkening, in the interim till a MaterialX implementation is done, which at least makes the darkening parameter do something. 

It seems much less disruptive to just change the MaterialX implementation later, than changing the model parametrization.

jstone-lucasfilm:
@portsmouth I'm not as concerned about the text of the specification changing after a v1.0 release, but we likely won't be able to change our reference implementation without a corresponding version number change, as this is the source document for the shading code that many teams (e.g. Karma, RenderMan, Omniverse) will use in their OpenPBR implementations.

portsmouth:
So how about we just ship with the interim MaterialX darkening solution borrowed from standard surface? This is easy to add.

jstone-lucasfilm:
@portsmouth I didn't get the sense that any artists liked the original Standard Surface coat darkening, and it seemed both visually unappealing and counterintuitive as a control system.  So that would likely be an unnecessary step backwards for the OpenPBR project overall.

portsmouth:
I mean it would have the same `coat_darkening` control, except adapted to turn on the Standard Surface coat darkening. You mean that that darkening effect is significantly worse than the OpenPBR one? It doesn't respect the IOR of the coat or base roughness, but for default IOR it should look plausible (just raises the base color to a power). 

I don't really agree with the logic of "let's release a less powerful model because implementation X doesn't have the time to ship version 1.234 by such and such date".

jstone-lucasfilm:
I'm hopeful that the new OpenPBR version looks better than the original Standard Surface coat darkening, which was virtually never used by artists, but I don't yet have a real-time implementation that I can test and evaluate!  The original Standard Surface version was pretty unappealing, though, from a visual and control system perspective.

portsmouth:
Or if there is not alternative, I will bite the bullet and implement the MaterialX graph for the darkening logic before the deadline (Probably some script to generate the XML is faster than attempting to do it my hand).

portsmouth:
>I'm hopeful that the new OpenPBR version looks better than the original Standard Surface coat darkening, which was virtually never used by artists, but I don't yet have a real-time implementation that I can test and evaluate! The original Standard Surface version was pretty unappealing, though, from a visual and control system perspective.

We did agree that it looks good verbally in a past meeting, if I recall. We seemed to agree that the darkening looks  natural when turned on, which is the default.

jstone-lucasfilm:
@portsmouth I'm not concerned about the time that it will take to **write** the coat darkening logic for OpenPBR in MaterialX, but rather the time that it will take to **validate** it visually and artistically, as one week is not sufficient time to show the functionality to a wide set of artists and get feedback.

jstone-lucasfilm:
It's worth noting that looking at a proposed visual feature in side-by-side renders is not the same as giving the control system to an artist and asking for their feedback!

portsmouth:
I  think this could be said of almost all the new features in the model. For example, surely the Zeltner sheen, thin film weight, fuzz on top, etc.  are not really validated by artists in that sense.  I don't think we can hope to achieve that by the 1.0 deadline,  for the coat darkening or anything else that changed since Standard Surface.

portsmouth:
>I've now merged https://github.com/AcademySoftwareFoundation/OpenPBR/pull/157, and the next step would be to resolve conflicts between these two pull requests, after which we can start on the MaterialX-side updates for this change.

So has the position changed since this, as seems like now you're pushing to omit the coat darkening from the 1.0?  (On the basis that we don't have time to validate it). We've spent quite a lot of time discussing and developing this darkening parametrization, for it to be thrown out at a late stage due to MaterialX concerns. I have also prototyped it in my viewer app, which has been available for months.

jstone-lucasfilm:
@portsmouth I'd strongly support the idea of including coat darkening for a 1.0 release in late June, but at our last meeting @AdrienHerubel specifically suggested that we omit it due to the new, shorter timetable for OpenPBR release.  I'd recommend bringing this up with Adrien!

portsmouth:
>It's worth noting that looking at a proposed visual feature in side-by-side renders is not the same as giving the control system to an artist and asking for their feedback!

The control system just turns the darkening on and off, so not much to it..  

NB, we also had Nikie in the meeting, where I recall she agreed the darkening looks good (though unfortunately we don't have the meetings recorded). So there's one artist validation data point, not sure how many we need to be confident.

I did point out on Slack that it's available to test in my viewer, but didn't get feedback on it (apart from when showing it in the meeting, where the feedback was positive as noted).

portsmouth:
It would be helpful if Adobe implemented/prototyped this in their renderer also.

jstone-lucasfilm:
I've had a chance to catch up with @portsmouth in a side discussion, and our current recommendation is to include the `coat_darkening` control in OpenPBR v1.0, but to assign it a default value of zero.

This will give us the flexibility to implement coat darkening using either BSDF node/layer upgrades or graph logic approximations in future versions of MaterialX, and in the meantime we'll have a visual match between our reference implementation and custom Arnold/Eclair shaders for the common case where `coat_darkening` is left at its default value.

In future versions of OpenPBR, we would then have the option of assigning a new default value of one to `coat_darkening`, with MaterialX upgrade logic used to transparently handle documents referencing the v1.0 version of OpenPBR.

portsmouth:
As per the commentary from @jstone-lucasfilm, I've made the change to have `coat_darkening` default to 0. While I think it's ultimately preferable for the physical darkening effect to happen by default, I think it's reasonable to have the effect be "opt-in" for now, until we have more confidence that it is preferable to change the default.

jstone-lucasfilm:
This looks good to me, thanks @portsmouth, and my only remaining question is whether we should add the new input to our reference implementation.

Since the coat darkening is not enabled by default, I wouldn't be as concerned about using a very simple reference implementation such as the color squaring of Standard Surface, if that seems like a reasonable compromise for now.

portsmouth:
> This looks good to me, thanks @portsmouth, and my only remaining question is whether we should add the new input to our reference implementation.
> 
> Since the coat darkening is not enabled by default, I wouldn't be as concerned about using a very simple reference implementation such as the color squaring of Standard Surface, if that seems like a reasonable compromise for now.

That seems reasonable, I will add it back. Though I'd note that if users decide they don't like the darkening effect on the basis of this, we need to bear in mind that the effect has not been properly implemented yet in MaterialX.

portsmouth:
@jstone-lucasfilm I added an approximate MaterialX implementation of the coat darkening math (based on the spec, not on the old Standard Surface formula). However, it seems the current MaterialX graph (even before the darkening addition) is failing to compile in the viewer, with error
```
Error in compiling fragment shader:
0(1723) : error C1503: undefined variable "base_substrate_out"
0(1724) : error C1503: undefined variable "base_substrate_out"
0(1798) : error C1503: undefined variable "base_substrate_out"
0(1799) : error C1503: undefined variable "base_substrate_out"
0(1878) : error C1503: undefined variable "base_substrate_out"
0(1879) : error C1503: undefined variable "base_substrate_out"
```
It's unclear what causes this, maybe it was introduced in a previous commit. Can you check and help to diagnose it?

jstone-lucasfilm:
Good catch, @portsmouth, and let me test the reference implementation in `main` to see what may have broken with respect to the v0.4 version.

jstone-lucasfilm:
Ok, I've tracked the issue down, and it was actually a limitation of GLSL generation in MaterialX 1.38, which has now been resolved by @niklasharrysson's refactor of the thin-film node in MaterialX 1.39.

Since we're actually targeting MaterialX 1.39 in OpenPBR, I would say that we're safe to move forward with both @virtualzavie's earlier change to thin-film controls and your current implementation of coat darkening, both of which render successfully in 1.39 builds.

In the near future, I'll add support for 1.39 features such as `color82` in our reference implementation, and then fully mark this as a MaterialX 1.39 document for clarity.

index.html
portsmouth: > > a blend between metal rm and dielectric rd roughness estimates
> 
> Readability might be improved by swapping the order of rm and rd to match the following formula:
> 
> > rb=lerp(rd,rm,M)

Done in e3cb364f95bd795afe3c61a08835f30ee3e8dedb

portsmouth: > > base roughness
> 
> Since `base_roughness` is the name of a parameter right now, it might be useful to use a different term than "base roughness" here (e.g., "effective base roughness" or something other than "base").

Done in a7f22fcddd71e944b5424df7a6c3a8618b303f71

jstone-lucasfilm:
For visual confirmation of your proposed change to the reference implementation, here's Anton's carpaint material with `coat_roughness` set to 1.0, and it looks very convincing to me:

image

jstone-lucasfilm (approved):
This looks great to me, thanks @portsmouth, and let's move forward with this update to OpenPBR.

Issue #142, opened 2023-11-14
Coat/specular IOR defaults, and TIR issue if coat IOR > spec IOR


portsmouth:
I think coat and specular IOR should be more different now, as the coat masks the specular if the IORs are the same. Current defaults are   coat=1.6, spec=1.5.

The coat IOR should probably be _lower_ than specular by default, to further boost the spec and prevent the TIR issue (e.g. coat=1.3, spec=1.6).

Also note that if the coat IOR is higher than the spec IOR, then TIR will occur in the specular reflection lobe. This looks weird if you don't account for the fact that the coat is sitting on top which bends the rays thus preventing the TIR, e.g. see the result boxed in red below.  (NB, the current Arnold and MaterialX implementations do this):

![image](https://github.com/AcademySoftwareFoundation/OpenPBR/assets/2774364/4b225977-98ae-451e-bed9-19e3b47d2a95)

Adobe proposes to avoid this by inverting the IOR of the base surface, though that seems to mess with the Fresnel physics which it would be best to avoid. (Though as a workaround in an implementation, maybe it's reasonable). We may want to note this problem and suggest some approaches, as otherwise a naive implementation could produce what looks like an artifact.

peterkutz:
> Adobe proposes to avoid this by inverting the IOR of the base surface, though that seems to mess with the Fresnel physics which it would be best to avoid.

The IOR inversion trick is physically inspired but not fully baked. I believe we can extend the trick of inverting the IOR to produce correct reflection coefficients regardless of the layer IORs. Here's a line of reasoning that leads to a new proposed approach:

1. Given a path consisting of a ray on one side of a surface and a refracted ray on the other side of the surface, the corresponding reflection coefficient at the surface is the same regardless of the direction of light flow along this path.
2. The reflection coefficient corresponding to a refraction event can be calculated on either side of the surface.
3. Only one of the two rays involved in the refraction event need to be known to correctly calculate the reflection coefficient.
4. The final angle of the light transmitted through a stack of layers only depends on the first and last IORs, not the IORs of the layers in between.
4. In the case of a coat medium sandwiched between an ambient medium and a base medium, the ray directions in the ambient and base media are fully specified by the ambient and base IORs (regardless of the coat medium IOR).
[...]

PR #166, opened 2024-02-26
Add a discussion/solution of the coat "spurious TIR" issue

portsmouth:
This addresses the points raised in https://github.com/AcademySoftwareFoundation/OpenPBR/issues/142.

I made it a sub-section of the Coat section, as it relates to accounting for the presence of the coat in the implementation of the dielectric/specular BSDF, but it could possibly go in the earlier Dielectric base section with a forward reference.

image


portsmouth:
An image of the artifact and the fix probably makes sense as well, to give some intuition.

index.html
```diff
@@ -852,10 +856,26 @@
 **`coat_affect_ior`**       | Affect IOR       | `float`  | $ [0, 1]   $ | $ 1         $ | How much to prevent the coat from affecting the relative IOR of layers below
  -->
 
-![](images/coat_0.png width=90% align=right) ![](images/coat_1.png width=90% align=left)
-<div class="shifted-caption">
-    ![Figure [coat]: Coat adds a secondary specular highlight and optional absorption tint](dummy)
-</div>
+
+### TIR
```
virtualzavie, line 860: Since it's only defined (and in any case it will always be defined after the table of contents):
```suggestion
### Total internal reflection
```

peterkutz:
This seems like a good place to put this discussion to me, since it's not very relevant before the coat is introduced.

peterkutz:
This is a nice addition and an interesting technique. However, to me doing these calculation at the microfacet level seems like overkill. I believe there's a simpler solution.

If I understand correctly, you are modifying the cosine at the microfacet level in order to squeeze the set of incident directions into a smaller range of angles, which avoids angles that result in TIR.

Instead of modifying the angles to squeeze them into the part of the Fresnel curve without TIR, we could equivalently modify the Fresnel curve to expand the part without TIR into the full range of angles.

Luckily, inverting an IOR results in the original Fresnel curve being squeezed or stretched into a smaller or larger set of angles, which respectively introduces or eliminates TIR. So we can achieve the desired outcome by inverting the relative IOR of the microfacet instead of modifying the cosine.

We can then take that one step further. Since all microfacets share the same relative IOR (unlike the cosine, which is different for each microfacet), we can simply modify the relative IOR of the whole base lobe, eliminating the need to modify the lobe internals.

This also avoids the costly second Fresnel evaluation for partial coats. For partial coats, we can simply blend the base IOR and the adjusted relative IORs (or the corresponding F0 values).

The implementation is straightforward and efficient: when $n_c < n_b$, simply set $\eta_\mathrm{bc}$ to $n_c/n_b$ (instead of $n_b/n_c$).

peterkutz:
I have one additional concern with the proposed approach, which is that modifying the cosine would also have the side effect of changing the metallic reflection from the microfacets as well (assuming that the dielectric and metal components share the same microfacet distribution). In particular, it would eliminate all grazing angles and thus eliminate the F82-tint edge color. This might be more physically correct for smooth base surfaces but it doesn't make sense for rough base surfaces where the microfacets could really be struck from any angle regardless of the coat refraction. Was this intentional?

jstone-lucasfilm:
I'm open to the simple inversion approach proposed by @peterkutz, and I believe this is similar to the approach taken in Adobe Standard Material.  What are your thoughts, @portsmouth?

peterkutz:
I have implemented the different options in Mathematica to verify that they work as expected.

The plots below show a dielectric configuration that exhibits TIR, the approach of inverting the relative IOR, and the approach of transforming the cosine. This illustrates that the IOR inversion and cosine transformation approaches have exactly the same effect.

OriginalSituationWithTIR

InvertedIORApproach

TransformedCosineApproach

(I'd be happy to run more tests or post the Mathematica notebook if useful.)

peterkutz:
I have implemented the different options in Mathematica to verify that they work as expected.

The plots below show a dielectric configuration that exhibits TIR, the approach of inverting the relative IOR, and the approach of transforming the cosine. This illustrates that the IOR inversion and cosine transformation approaches have exactly the same effect.

(I'd be happy to run more tests or post the Mathematica notebook if useful.)

OriginalSituationWithTIR
InvertedIORApproach
TransformedCosineApproach

portsmouth:
I think there may be a slight confusion about formula I proposed.  In my proposal I had:

$$
\mu_c^2 = 1 - (1 - \mu_i^2) / \eta^2_\mathrm{ca} \ .
$$

Which simply gives the angle of the ray inside the coat, after it is refracted from the exterior ambient medium into the coat. Note that it is $\eta_\mathrm{ca}$ that appears here, *not* $\eta_\mathrm{bc}$ (as this is the refraction at the coat/ambient boundary). I then required that this angle is the one we use in the microfacet Fresnel formula for the base (specular) lobe:

$$
F(\mu_c, \eta_\mathrm{bc})
$$

where in this Fresnel factor, $\eta_\mathrm{bc}$ is used since this ray is reflecting off the coat/base boundary. Since physically, if the coat was smooth, that is the actual angle of incidence at the coat/base boundary. (The coat is not generally smooth, so this is approximate, but we avoid the TIR artifact at least).

------------

In your formula, I take it you are replacing that Fresnel factor with:

$$
F(\mu_i, \eta_\mathrm{cb})
$$

i.e. just inverting the IOR ratio at the coat/base boundary. The plots you show are showing that this equals a Fresnel factor with the inverse IOR ratio:

$$
 F(\mu^\prime_i, \eta_\mathrm{bc}) = F(\mu_i, \eta_\mathrm{cb})
$$

(That follows basically from time-reversal invariance I think, i.e. the Fresnel factor for A -> B is the same as that for B -> A, if you imagine the rays reversing direction). But  I think the $\mu^\prime_i$ that satisfies the equation above is the direction of the $\mu_i$ ray if refracted at the coat/base boundary, *not* $\mu_c$ which is what we wanted to match.

So I see that your approach avoids the TIR, since it artificially swaps the IOR of coat and base so that the coat IOR is always lower. But I don't think that is faithful to the physics, as this doesn't correspond to any actual Fresnel factor in the true light transport, even in the smooth limit.

Of course, I could be misunderstanding or missing something in my thinking about your proposal.

----------------

>I have one additional concern with the proposed approach, which is that modifying the cosine would also have the side effect of changing the metallic reflection from the microfacets as well (assuming that the dielectric and metal components share the same microfacet distribution). In particular, it would eliminate all grazing angles and thus eliminate the F82-tint edge color. This might be more physically correct for smooth base surfaces but it doesn't make sense for rough base surfaces where the microfacets could really be struck from any angle regardless of the coat refraction. Was this intentional?

True, though this seems hard to avoid without doing a full Monte Carlo simulation of the scattering between the rough interfaces. If there is a coat over a metal, won't the coat Fresnel generally mask the F82 tint anyway? It could be useful to study some renders to see if there really is a significant artifact introduced by my form of the TIR fix.

peterkutz:
> But I think the that satisfies the equation above is the direction of the ray if refracted at the coat/base boundary, not which is what we wanted to match.

Good point. After reading your replies and looking at the problem more closely, I see that while my solution produces a plausible result, it does not produce the correct results in general. Instead, it produces a base specular reflection that is slightly too intense (although in practice this might actually help to make up for the energy that's missing due to ignoring coat internal reflections).

Instead of inverting the IOR, I wonder if there would be a way to transform the IOR to produce the exact result or at least a more accurate result. Since you're good at this type of mathematical problem, I wonder if you might have an idea of how to do this.

At a high level, I still prefer the approach of modifying the IOR for a few reasons:

- The IOR can be modified once and then provided to the base specular lobe.
- The internals of the base specular lobe do not need to be modified at all.
- There is no confusion about whether and when to modify incoming and outgoing angles.
- BSDF evaluation, BSDF sampling, and light sampling can proceed as usual once the IOR is modified.
- The slightly excessive base specular reflection might help compensate for the missing outgoing light due to the lack of internal reflections.
- Both solutions are approximations anyway and modifying the IOR seems easier to describe and implement.

Even if the IOR inversion technique can't easily be made more accurate, I would propose noting it as an alternative solution for implementers that find the cosine-modification technique too invasive or complex.

Below are a couple relevant attachments.

The Mathematica code (in text form because .nb files can't be attached) used to generate the above plots (which are relevant but not fully representative of the actual situation in question):

[FresnelComparisonsForSpuriousTIRIssue.txt](https://github.com/AcademySoftwareFoundation/OpenPBR/files/15100227/FresnelComparisonsForSpuriousTIRIssue.txt)

The test renders I did for ASM validating the plausibility and reasonable accuracy of the IOR inversion technique (from the [ASM technical document](https://helpx.adobe.com/substance-3d-general/adobe-standard-material/asm-technical-documentation.html)):

ASMSolutionForSpuriousTIR

peterkutz:
I have also created some slides to intuitively illustrate how the IOR inversion technique works as well as its limitations. (It's best to flip between the pages like a slideshow to more easily see what changes from slide to slide.)

[PeterProposedFixForSpuriousTIR.pdf](https://github.com/AcademySoftwareFoundation/OpenPBR/files/15100311/PeterProposedFixForSpuriousTIR.pdf)

portsmouth:
@peterkutz Regarding your critique above, what I meant (but didn't write up correctly), is that the diagram below is supposed to be in the local space of the base microsurface,  and in the approximation that the over-lying coat microsurface has matching micronormal.  I realize this is not strictly correct (i.e. the coat micronormal of course may differ), but for the purposes of the Fresnel calculation it's an (imperfect) approximation that doesn't introduce any major artifact but fixes the major TIR artifact.

So the $\mu_i$ in this picture is $\omega_i \cdot m$ where $m$ is the shared micronormal. From that you can compute $\mu_c$ and the Fresnel factor $F(\mu_c, \eta_\mathrm{bc})$, which avoids the TIR.

image

I don't intuitively see how inverting the base IOR (or otherwise changing it) is a better approximation, since this avoids the TIR by essentially fudging the IORs in the calculation to make the TIR impossible. Though it is simpler to implement I admit. Perhaps we can compromise by you adding a brief explanation of the TIR inversion trick as an additional suggestion?

(And I can also fix up my proposal to make the assumptions about the microsurface more clear).

portsmouth:
> Instead of modifying the angles to squeeze them into the part of the Fresnel curve without TIR, we could equivalently modify the Fresnel curve to expand the part without TIR into the full range of angles.
> We can then take that one step further. Since all microfacets share the same relative IOR (unlike the cosine, which is different for each microfacet), we can simply modify the relative IOR of the whole base lobe, eliminating the need to modify the lobe internals at all.
> This also avoids the costly second Fresnel evaluation for partial coats. For partial coats, we can simply blend the base IOR and the adjusted relative IOR (or the corresponding F0 values) and provide that to the base specular lobe.
> The implementation is straightforward and efficient: when nc>nb, simply set ηbc to nc/nb (instead of nb/nc).

To be fair, this all sounds highly plausible and could be more efficient than what I propose. So perhaps what we can do is refactor my discussion to state the *source* of the problem (the bending of the ray),  but state your scheme as an approximate and efficient solution.  Do you mind to attempt to draft that?

peterkutz:
Thanks for the clarifications @portsmouth . It makes more sense and if I understand correctly it is just trying to approximate the actual physics where the light is reaching the base surface from a limited cone of directions. It seems fine to include (the clarified version of) this suggestion in the spec.

> fudging the IORs ... simpler to implement ... TIR inversion trick as an additional suggestion

I think that might be good to include the IOR modification as an additional suggestion since they are both approximations anyway (compared to something like the Weidlich-Wilkie approach) and since the IOR modification might be easier to implement and better than nothing.

> Do you mind to attempt to draft that?

Sure, I wouldn't mind drafting something, although I won't have availability to do that until after the 1.0 release.

Does anything special need to be done for light samples with your approach?

portsmouth:
>Does anything special need to be done for light samples with your approach?

Not that I'm aware.

portsmouth:
I think this is good to merge now, pending approval from Julien and/or Peter.

images/dielectric_tir_issue.svg
virtualzavie, line 1: Strictly speaking, the entering and exiting angles are not the same.

virtualzavie (approved):
Yes, I think that's a good to merge.
Trying to evaluate the exact behaviour is probably ridiculously complex, but the PR poposes two reasonable approximations to pick from. 👍

images/dielectric_tir_issue.svg
portsmouth: You mean in the diagram? I can try to fix it..

jstone-lucasfilm (approved):

Issue #150, opened 2023-12-28
Thin film in micrometers divided by 2 for normalized parameter range


sharktacos:
Currently thin film has a range of 0-2000 nanometers. I'd like to propose for discussion instead measuring in micrometers which would be 0-2 range, and then dividing that by 2 in order to have a normalized 0-1 parameter range. This makes a more artist friendly slider.

portsmouth:
I like this, the existing requirement to input a length in nm seems a bit obscure to me.

portsmouth:
Also, it's unclear from the spec how the `thin_film_ior` interacts with the IORs of the surrounding media. Should the thin-film disappear if its IOR equals the coat IOR for example? (Physically, presumably yes, but in implementations I've seen it currently doesn't).

portsmouth:
A related proposal is to add a `thin_film_weight` parameter in [0,1], which will just blend the effect in and out. If we had that, keeping the width parameter in physical units might be more reasonable.

https://github.com/AcademySoftwareFoundation/OpenPBR/issues/134

jstone-lucasfilm:
Thanks for this original report, @sharktacos, and it's now addressed in #167.

Issue #134, opened 2023-10-30
For consistency, every section should have a "weight" parameter.


ZapAndersson:
While working on the UI for 3ds max, I realize how inconsistent it feels that everything has a weight, _except_ emission and thin film.

I think _everything should have a weight_, even though this is just a multiplier in the case of emission and thin film, the UI consistency and the ease of explaining the functionality wins out, IMHO.

Also, since both thin-film thickness and emission luminance quite often have values outside the easily texturable 0-1 range, it is _extremely convenient_ to be able to set e.g. a high luminance value, but then modulate the intensity of the light by a 0-1 weight map, or to set a given thin-film thickness, and modulate the coverage of the thin film with a 0-1 weight map.

The only reason not to have them is parameter count frugality, but if most real-world use cases has users plugging in multiplier nodes in front of these all the time, I feel that feature should be built in.

andre-ilm:
I'd like to hear more about the reasoning for weights everywhere. For our internal proprietary shaders we took the opposite approach for a few reasons:
- Fewer maps (can be critical for GPUs)
- Singular controls (don't have more than one way to do the same thing)
- An external shader graph can provide this control if necessary

portsmouth:
After https://github.com/AcademySoftwareFoundation/OpenPBR/pull/231 this issue has been resolved. @ZapAndersson please close.

Issue #79, opened 2023-08-08
Implementing thin-film interference in combination with F82-tint metallic Fresnel


lukasstockner:
The current spec draft points out [the method used for implementing thin-film interference effects](https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html). However, this paper and its reference implementation assume that the physically correct dielectric and conductive Fresnel terms are used so that the optical phase shift can be computed.

However, OpenPBR uses the F82-tint model for metals, which directly computes the reflectivity without going through computing a complex IOR and then using it in the conductive Fresnel term.

Therefore, it might be helpful to include a note in the spec on how to reconcile this and how to compute/estimate the phase shift when implementing the thin-film component.

portsmouth:
We need a PR to clarify the thin film behavior.

jstone-lucasfilm:
Thanks for this original report, @lukasstockner, and it's now addressed in #167.

PR #167, opened 2024-02-28
Thin-film improvements

portsmouth:
This addresses:
  - https://github.com/AcademySoftwareFoundation/OpenPBR/issues/79
  - https://github.com/AcademySoftwareFoundation/OpenPBR/issues/134
  - https://github.com/AcademySoftwareFoundation/OpenPBR/issues/150

We still don't give much detail about the implementation, but at least we make the physics that is supposed to be implemented reasonably clear -- or as clear as it can be given our usage of the not-strictly-physical F82-tint conductor model:

>Modeling of the precise effect is implementation-dependent, however as described in the Metal section the Fresnel factor for metal is defined according to the Schlick-based "F82-tint" parametrization, which does not specify the underlying (complex) physical IOR. We suggest here that some reasonable approximation is employed to map the Fresnel factor to effective IOR and absorption coefficients, for example that described by [#Gulbrandsen2014].

As discussed on Slack and in the tickets above, it makes a great deal of sense to add a `thin_film_weight` parameter. This improves the expressivity of the model (in some use-cases, it could be essential to smoothly blend the effect in), and is a fairly trivial change to the implementation. 

Also in a sense it didn't make sense to have the default `thin_film_thickness = 0`, as in fact the effect doesn't smoothly turn off as thickness goes to zero, it causes darkening for any finite thickness. I suggest instead we change the default `thin_film_thickness` to e.g. 1000 (mid-range), and have the `thin_film_weight` default to 0. Then on dialing this weight parameter the effect turns on smoothly. We may wish to tweak the default from 1000 though, depending on the look.

image

image

It seems better to break the thin-film discussion out into its own sub-section of the "Base Substrate" section.


portsmouth:
Another issue to contend with is that currently the default `thin_film_ior` and `specular_ior` are both 1.5. According to the physical description, this means that the thin-film would have no effect if `metalness=0`, since it is index-matched with the underlying dielectric. Similarly, the fringes should also disappear if the coat is present but index-matched with the film.

An implementation could ignore the IOR of the underlying dielectric (e.g. Arnold does, currently), but this would technically be an incorrect interpretation of the physics. Also of course the actual fringes generated by the model would be fake as they should change as the IOR of the base changes.

Instead, I suggest we tweak the default IORs so that the thin-film fringes appear (to some reasonable extent) when used with those defaults. The thin-film IOR default simply has to be sufficiently different to both the specular and coat IORs. So perhaps e.g.:
 - `specular_ior`=1.6 ($\approx$ polycarbonate)
 -  `thin_film`=1.4 ($\approx$ oil)
 - `coat_ior`=1.3  ($\approx$ water)

jstone-lucasfilm:
@portsmouth Would it make sense to update our reference implementation as part of this change, so that the specification and implementation remain in sync?

index.html
```diff
@@ -446,17 +446,25 @@
...
+### Thin-film iridescence
 
-!!! Note: Future work
-   It is currently not physically clear enough how the IOR of the surrounding media affects the thin-film interference, and how the thin-film affects the Fresnel of the adjacent slabs. Moreover, the behavior of the current parametrization is arguably unintuitive, as the effect of the thickness is not obvious. It may be useful to have a more artist-friendly dial of the "iridescence".
+_Iridescence_ is the occurrence of rainbow-like color fringes when a thin dielectric film with thickness on the order of the wavelength of light is placed on top of a material, due to wave interference between the various electromagnetic reflection modes within the film. To model this, there is assumed to be such a thin-film sitting on top of the base substrate (whether metal or dielectric), parametrized only by:
+
+   - the coverage (presence) weight of the film, **`thin_film_weight`**,
+   - the thickness of the film in nanometers, **`thin_film_thickness`**,
+   - the index of refraction (IOR) of the film, **`thin_film_ior`**,
```
virtualzavie, line 455:
```suggestion
   - the index of refraction (IOR) of the film, **`thin_film_ior`**.
```

index.html
```diff
@@ -446,17 +446,25 @@
...
+
+In practice, this wave-optics effect is most easily incorporated directly into the Fresnel factor of the microfacet BSDFs of both the metal and dielectric-base layers. The currently recommended thin-film model is that of Belcour and Barla [#Belcour2017]. The coverage weight functions as a blend between the BSDF with and without the presence of the film, and is necessary since neither the thickness nor the IOR of the film provide a dial of the effect without altering the shape and saturation of the color fringes.
```
virtualzavie, line 457: I don't think it's necessary to defend the parametrisation, so I propose to simply say:
```suggestion
In practice, this wave-optics effect is most easily incorporated directly into the Fresnel factor of the microfacet BSDFs of both the metal and dielectric-base layers. The currently recommended thin-film model is that of Belcour and Barla [#Belcour2017]. The coverage weight functions as a blend between the BSDF with and without the presence of the film, and thus allows to dial the effect without altering the shape and saturation of the color fringes.
```

index.html
```diff
@@ -446,17 +446,25 @@
...
 
+The interference within and resulting reflection from the film will be affected (as described by Belcour and Barla [#Belcour2017]) by the complex IOR of the adjacent media above and below the film, which in general are a statistical mix of metal and dielectric below and coat and ambient medium above. Modeling of the precise effect is implementation-dependent, however as described in the Metal section the Fresnel factor for metal is defined according to the Schlick-based "F82-tint" parametrization, which does not specify the underlying (complex) physical IOR. We suggest here that some reasonable approximation is employed to map the Fresnel factor to effective IOR and absorption coefficients, for example that described by [#Gulbrandsen2014].
```
virtualzavie, line 459: I don't understand:
> The interference within and resulting reflection from the film will be affected [...]

I'm not sure wether it's meant as an ellipsis (_within [the layer]_), or if it's a typo. I think it should be more explicit.
Also, the sentence is fairly long, so I suggest to split it.

index.html
```diff
@@ -446,17 +446,25 @@
...
---------------------------|-----------|----------|:---------------:|:-------------:|:-------:|----------------------------------------------
-**`thin_film_thickness`** | Thickness | `float`  | $ [0, \infty) $ | $ [0, 2000] $ | $ 0   $ | Thickness of the film in nanometers
-**`thin_film_ior`**       | IOR       | `float`  | $ (0, \infty) $ | $ [1, 3]    $ | $ 1.5 $ | Refractive index of the film
+
+
+Thin-film params          | Label     | Type     | Range           | Norm          | Default  | Description
+--------------------------|-----------|----------|:---------------:|:-------------:|:--------:|----------------------------------------------
+**`thin_film_weight`**    | Weight    | `float`  | $ [0, 1]      $ |               | $ 0    $ | Coverage weight of the thin film
+**`thin_film_thickness`** | Thickness | `float`  | $ [0, \infty) $ | $ [0, 2000] $ | $ 1000 $ | Thickness of the film in nanometers
+**`thin_film_ior`**       | IOR       | `float`  | $ (0, \infty) $ | $ [1, 3]    $ | $ 1.5 $  | Refractive index of the film
```
virtualzavie, line 467: I like @sharktacos's proposal (#150) to have a normalised parameter. This can be done in a separate PR though if there isn't consensus there.

virtualzavie (commented):
> I suggest instead we change the default thin_film_thickness to e.g. 1000 (mid-range), and have the thin_film_weight default to 0.

I agree. Parameters should have reasonable default values. To me this is like choosing a default roughness for the coat or a default colour for the subsurface scattering.

index.html
portsmouth: Agreed, though note that "allows to" should be modified to "allows one to" to be grammatical. ("Allows" needs an object which is being allowed to do the thing).

index.html
portsmouth: It's not a typo, I was alluding to the fact that the color fringes arise from interference of the ladder of reflection modes *within* the film,  the sum total effect of which is the color-fringy appearance of the reflection.

If you think it sounds confusing, let's reword. Perhaps:
>The shape and color of the fringe patterns in the reflection from the film will be affected (as described by Belcour and Barla) by the complex IOR of the adjacent media ...

index.html
portsmouth: I think you could probably get away making the width be in micrometers, with a soft-max at 1.  So have it default to [0,1] range.

In the Belcour & Barla paper, they use the parameter $\mathcal{D} = 2 \eta d$ where $\eta$ is IOR, $d$ is width. In their tests, $\mathcal{D}$ seems to vary in range [0, 3200],  though with maximal effect at the mid-range i.e. 1600nm. So that implies even if $d$ is (soft) capped at 1 micrometer, the useful range is captured.

virtualzavie: Yup, sounds good.

index.html
virtualzavie: Thanks, I've updated the suggestion.

index.html
virtualzavie: Yes, I find this new version much easier to read.

portsmouth: Committed in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/167/commits/abf2f7d2cdce3b78de5e284e97dcfc93e7902ad2.

portsmouth:
> @portsmouth Would it make sense to update our reference implementation as part of this change, so that the specification and implementation remain in sync?

Done in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/167/commits/f8538ece5a5c89c21e6c2096131ae5ce320ad9f1.

virtualzavie (approved):
LGTM

portsmouth:
What about this:

https://github.com/AcademySoftwareFoundation/OpenPBR/pull/167#issuecomment-1969900347

Should we change the default IORs? As currently the thin-film will be invisible if `base_metalness=0`, since its IOR is equal to the dielectric IOR.

AdrienHerubel:
Let's move the default value changes to a separate PR and discuss there, let's go ahead with the unit change.

portsmouth:
In https://github.com/AcademySoftwareFoundation/OpenPBR/pull/167/commits/984bff1ccd08444dbd6707ee079c861b62deb82d I added the units change. 

I also added more detail to the text regarding the implementation, and a clarifying diagram based on the one above.

I've refactored the section layout also, to make the thin-film discussion separate from the "Base substrate" section. This seems to make more logical sense, since now we cover the entire base (metal and dielectric slabs), *then* the thin film which sits on top of it, *then* the coat, *then* the fuzz.

image

image

index.html
```diff
@@ -751,56 +782,38 @@
...
-Metal
+Thin-film iridescence
 -------------------------------------
 
-Metals are completely opaque and have a characteristic and familiar form of specularity due to the Fresnel factor for conductors differing from that for dielectrics.
-The metallic base is thus represented as a separate bulk slab which consists of an opaque GGX microfacet conductor BRDF $f_\mathrm{conductor}$, whose NDF is parametrized by **`specular_roughness`**, **`specular_anisotropy`**, and **`specular_rotation`** (overloading the same parameters used for the dielectric BSDF $f_\mathrm{dielectric}$, as noted in the [Microfacet model](index.html#model/microfacetmodel) section):
-\begin{equation}
-S_\mathrm{metal} = \mathrm{Slab}(f_\mathrm{conductor}) \ .
-\end{equation}
+_Iridescence_ is the occurrence of rainbow-like color fringes in the reflection when a thin dielectric film with thickness on the order of the wavelength of light is placed on top of a material, due to wave interference between the various electromagnetic reflection modes within the film. To model this, there is assumed to be such a thin-film sitting on top of the base substrate (whether metal or dielectric), parametrized only by:
```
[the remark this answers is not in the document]
portsmouth: That seems likely to be a special case deserving of a specific model, but probably thin-film iridescence is a reasonable fake approximation to it at least.

portsmouth:
Does this sound dumb?

>In principle the implementation should model all these configurations correctly, though modeling of the precise effect is implementation-dependent.

index.html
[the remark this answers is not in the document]
virtualzavie: Birds seem a common enough case to me to deserve a bit of attention. I'm making a mental note to test whether the thin-film model is sufficient for that case.

index.html
```diff
@@ -545,22 +545,23 @@
 \begin{equation}
 \mathbf{F}_{82}(\mu) = \mathbf{F}_{\mathrm{Schlick}}(\mu) - \frac{\mu (1 - \mu)^6}{\bar{\mu}(1 - \bar{\mu})^6} \Bigl(\mathbf{F}_{\mathrm{Schlick}}(\bar{\mu}) - \mathbf{F}(\bar{\mu})\Bigr)
 \end{equation}
-where $\bar{\mu} = 1/7$, and $\mathbf{F}(\bar{\mu})$ is the desired metallic reflectivity at that "grazing edge" angle cosine corresponding roughly to $82^\circ$ (i.e. around silhouettes), ensuring $\mathbf{F}_{82}(\bar{\mu}) = \mathbf{F}(\bar{\mu})$. This desired edge reflectivity is user-specified as a fractional tint of the Schlick curve, i.e.
+where $\bar{\mu} = 1/7$, and $\mathbf{F}(\bar{\mu})$ is the desired metallic reflectivity at that "grazing edge" angle cosine corresponding roughly to $82^\circ$ (i.e. around silhouettes), ensuring $\mathbf{F}_{82}(\bar{\mu}) = \mathbf{F}(\bar{\mu})$. This desired edge reflectivity is user-specified as a fractional tint of the Schlick curve control via **`specular_color`**, i.e.
```
virtualzavie, line 548: Did you mean "tint of the Schlick curve control, via specular_color" or "tint of the Schlick curve, controlled via specular_color"?

portsmouth: Yes should be the latter, good catch!

virtualzavie:
Following the reply from Jamie, I'm proposing a minor change.
Other than that, the PR looks good to me and ready to merge.

virtualzavie:
Addendum: the `open_pbr_pearl.mtlx`, `open_pbr_soapbubble.mtlx` and `open_pbr_surface_default.mtlx` files need to be updated as the parametrisation have changed.

portsmouth:
>Addendum: the open_pbr_pearl.mtlx, open_pbr_soapbubble.mtlx and open_pbr_surface_default.mtlx files need to be updated as the parametrisation have changed.

Done in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/167/commits/ac476eb98263c9073f44b520461f6704bfb8e5f1.

virtualzavie:
All looks good to me and I'd say the PR is ready to merge. 👍

jstone-lucasfilm (approved):
This looks good to me as well, thanks @portsmouth and @virtualzavie!

portsmouth:
It seems the .svg file fails to render in the live version of the spec, though it did in my local browser for some reason.

Issue #110, opened 2023-10-08
Clarify handling of rays incident to the surface from the interior


portsmouth:
Currently we don't say much about how the model deals with the difference between rays exiting and entering the surface. This has to be handled in a renderer (at least one which tries to correctly render glass objects with a coat/fuzz, for example) so we should clarify this. 

For example consider the case of a glass object, with a coat and fuzz. Rays entering from the exterior (i.e. the ambient dielectric medium) will enter through the fuzz, then the coat, then transmit into the base glass.

Rays which hit the surface from the interior of the glass (having refracted into the glass at some earlier point in the path) instead hit the _bottom_ side of the coat, then the fuzz, then transmit into the ambient medium:

![image](https://github.com/AcademySoftwareFoundation/OpenPBR/assets/2774364/830ba2d3-9fb0-408a-a497-171f028ad79a)

The physical effect of the layers differs in these case. For entering rays:

   - the fuzz reflection is un-tinted by the coat absorption
   - the coat reflection is dimmed and roughened by the fuzz
   - the dielectric reflection and transmission are both dimmed and roughened by both the coat and the fuzz

While for rays which are exiting:
   - the fuzz reflection (viewed from inside) now _will_ be tinted by the coat absorption
   - the reflection from the coat "top" interface (with the fuzz) is _not_ dimmed or roughened by the fuzz. This reflection is also an internal one, so has different Fresnel factor.  
   - only dielectric transmission is dimmed/roughened by coat and fuzz, while the reflection is unaffected (except for being an internal reflection as well)
   
However in Standard Surface (and MaterialX), we make no attempt to account for this in our albedo-scaling approximation of the layering. The Arnold implementation currently uses a rather crude approach where the normal is flipped so the surface always thinks the ray is entering 🤦‍♂️  (Then we have to use some messy logic to make nested dielectrics work despite this).

In reality there is kind of a symmetry between entering and exiting. In both cases the ray transmits from one external dielectric medium to another, via some intervening interfaces and layers of media, which we know. A sufficiently powerful layering formalism/system should be able to compute (within some reasonable approximation) the BSDF accounting correctly for this. It is possible to write down an albedo scaling approximation of this which looks symmetric, for example (which I didn't attempt in the spec). 

It is probably too much detail for us to elaborate on this in the spec (at least at this stage), but I think we should at least discuss what would need to be done to correctly model the physics. 

(Note, @iliyang and I discussed this issue previously in relation to Standard Surface).

[...]

PR #168, opened 2024-03-04
Add a discussion of the difference between entering/exiting light transport

portsmouth:
Based on https://github.com/AcademySoftwareFoundation/OpenPBR/issues/110.

I added this discussion as a sub-section of the "Reduction to a mixture of lobes" section, as it is also quite technical and seems to make the most sense in the context of trying to do an implementation of the layering.

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -1203,7 +1203,7 @@ The full representation in code of this linear combination of BRDF/BTDF/BSSRDF l
 
 ### Entering versus exiting
 
-As aspect that has been ignored in this approximation (in the non-thin-walled case) is the different light transport for rays incident from the top or the bottom side. In reality the physical effect of the layers differs in these cases. For entering rays:
+An aspect that has been ignored in this approximation (in the non-thin-walled case) is the different light transport for rays incident from the top or the bottom side. In reality the physical effect of the layers differs in these cases. For entering rays:
 
    - The fuzz reflection is un-tinted by the coat absorption.
    - The coat reflection is dimmed and roughened by the fuzz.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1197,6 +1201,22 @@
 
 The full representation in code of this linear combination of BRDF/BTDF/BSSRDF lobes will depend on renderer-specific details.
 
+### Entering versus exiting
+
+As aspect that has been ignored in this approximation (in the non-thin-walled case) is the different light transport for rays incident from the top or the bottom side. In reality the physical effect of the layers differs in these cases. For entering rays:
```
virtualzavie, line 1206:
```suggestion
An aspect that has been ignored in this approximation (in the non-thin-walled case) is the different light transport for rays incident from the top or the bottom side. In reality the physical effect of the layers differs in these cases. For entering rays:
```

index.html
```diff
@@ -1197,6 +1201,22 @@
...
+
+   - The fuzz reflection is un-tinted by the coat absorption.
```
virtualzavie, line 1208:
```suggestion
   - The fuzz reflection is not tinted by the coat absorption.
```
("un-tinted" sounds to me like a tint is removed somehow)

virtualzavie (approved):
Looks fair to me.
I've let two minor changes.

peterkutz (approved):
This sounds good to me. The guidelines are intuitive and reasonable, and the description seems to have the right level of detail.

jstone-lucasfilm (approved):

Issue #155, opened 2024-01-23
Anisotropy direction parametrization


virtualzavie:
The current OpenPBR specification proposes to control the anisotropy with a `specular_anisotropy` (respectively `coat_anisotropy`) and the direction of the specular highlight elongation with a `specular_rotation` (respectively `coat_rotation`).

In the case those inputs are given with a texture, or more generally if some filtering is involved, the `specular_rotation` will require special care to avoid artifacts due to the discontinuity at the 360° / 0° angle. At a minimum, this requires the ability to specify a nearest-neighbor filter, but obtaining a level of quality consistent with bilinear, trilinear or anisotropic filtering involves implementing a custom filtering scheme to handle the discontinuity case. Such a custom filtering has to be implemented deep into the shader, requires more texture fetches, and represents a non trivial amount of work for the developer of the renderer.

An alternative approach is to specify the direction with a flow map of the anisotropy direction 2D vectors. This parametrization is more suitable to texture filtering implemented in most renderers and 3D hardware, does not require a special case, and is very similar to normal maps which are widely supported.

Although it may seem expressing an angle (1 parameter) as a vector (2 parameters) might incur an additional bandwidth cost, factoring `specular_anisotropy` into the vector norm should lead to an equivalent cost (or lesser since no custom filtering with additional fetches is involved).

[glTF expresses the anisotropy direction and strength as a 3 component texture](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_anisotropy) though, so pro and cons of the two solutions would have to be weighted.

It may also seem that authoring a flow map directly is more difficult, but authoring a rotation map directly is in fact difficult as well and better performed with a dedicated tool anyway.

portsmouth:
It would be useful to see some renders. Is it not possible to avoid this discontinuity by computing the tangent vector before the interpolation?  (That is, blend vectors, not angles).

virtualzavie:
> It would be useful to see some renders.

As an illustration, here is the same model:
- displayed correctly, thanks to a custom filtering implemented in the fragment shader:
<img width="349" alt="2024-01-30 Correct anisotropy angle filtering" src="https://github.com/AcademySoftwareFoundation/OpenPBR/assets/9511025/b30f08d5-138c-41e2-9ac1-184284b2ce07">

- displayed with a MIP mapping bias such that the custom texturing filtering no longer prevents the anisotropy angle discontinuity from causing a visible seam:
[...]

PR #170, opened 2024-03-19
Specify the anisotropy direction

virtualzavie:
This PR removes entirely the definitions ~~`specular_rotation` and `coat_rotation`~~, to consider instead that the provided `geometry_tangent` is sufficient to define the direction of the anisotropy.

To avoid correlation between the NDF of the base layer and coat layer, `geometry_coat_tangent` is added.

Finally the Normal maps section is developed to mention how a tangent map can alter the tangent vector in the same way a normal map alters the normal vector.

image
image
image
image
image
image
image
image


virtualzavie:
I've added several changes, as proposed in #155:

Naming ([d7d7040](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/170/commits/d7d70409fdd51b6d565a5df6f15df360c8849732)):
image
image
image
image

Specification of the direction as a property of the NDF ([e321ff9](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/170/commits/e321ff94a521d78eb9d2f08c5f94d59d50dfe852)):
image
image
image
image
image

portsmouth:
I feel strongly that we should keep the existing `specular_rotation` and `coat_rotation` parameters, applied as an additional angular offset on top of the flow-map generated T, B. This should normally only be used to globally rotate the highlights (but in some cases could be textured, with the caveat about the interpolation artifacts due to the discontinuity which we should warn about in the spec). This ability to globally rotate seems useful to retain. 

Also, there will be existing assets that texture the rotation. A case where it could genuinely make sense to texture the rotation would to add some perturbing angular noise to procedurally break up the flow-map generated directions (which will not introduce any artifact provided the noise is smooth).

portsmouth:
Regarding the discussion in the meeting about what assumptions we should make about the tangent space, i.e. assume the policy is roughly:
  - a reference tangent space (N, T, B) is assumed to be known. We do provide the parameters `geometry_normal` and `geometry_tangent` which are supposed to specify this, though I think in practice the renderer itself will normally ensure the shader knows the local tangent frame, adjusted according to the shading normals and tangents supplied with the geometry, modulated by any normal and displacement maps.
  - the final tangent frame is defined relative to this reference space. I think we should probably write this formula down in the spec, to avoid any confusion about how this is defined
 
image

virtualzavie:
Yes, good point. We already state the assumption of the existing frame of reference, but the exact transform is certainly lacking.

portsmouth:
>The direction is specified as a normalized 2D vector, with $(1, 0)$ being the original tangent vector direction.

I think we need to be more explicit that the values in the texture, $(R, G)$ (where $R, G \in [0,1]$ after dequantization), are mapped to $(t, b)$ via:

$$
t = 2R-1
$$

$$
b = 2G - 1
$$

where we require that $t^2 + b^2 =1$. Then the final tangent vector is given (in terms of the reference tangent frame) by:

$$
\mathbf{T}' = t \mathbf{T} + b \mathbf{B}
$$

virtualzavie:
There were several points brought, so I'll try to address them one by one:

- On the additional rotation parameter: I am not convinced this value is worth the cost. This reintroduces the implementation complexity that the flowmap was meant to avoid. It also adds a parameter, which is something we should try to limit, for a control that's essentially redundant with the anisotropy direction. I hear the use case, but that sounds like a proposal to work around a potential limitation of the authoring tool. I would instead suggest to leave this proposed parameter out at least until we have a clear demand for it.

- On the specification of the tangent space: agreed, will amend the PR to specify the relation between $T$ and $T'$.

- On the direction of the grooves vs direction of the highlight: your comment makes me realise I confused what the effect of stretching the NDF would be. I will have to amend the PR to fix that. As for $T$ vs $B$, I agree it makes more sense for the artist to draw the direction of the groves, but as you point out, the vector field will express both equivalently. So in the end I think we can conclude it's irrelevant, which means the decision to how best express it will be left to the authoring tool. On the specification side, we then should probably choose the simplest and clearest specification. I think "NDF stretching along T" would be it (even though it means the opposite of what I originally thought it meant 🙃).

- On the mapping from $[0, 1]$ to $[-1, 1]$: [I was asking](https://github.com/AcademySoftwareFoundation/OpenPBR/issues/155#issuecomment-2009614171) because I was not sure there was a consensus, but I'll update the PR to reflect this.

Did I forget anything?

portsmouth:
>On the additional rotation parameter: I am not convinced this value is worth the cost. This reintroduces the implementation complexity that the flowmap was meant to avoid. It also adds a parameter, which is something we should try to limit, for a control that's essentially redundant with the anisotropy direction. I hear the use case, but that sounds like a proposal to work around a potential limitation of the authoring tool. I would instead suggest to leave this proposed parameter out at least until we have a clear demand for it.

Sounds reasonable. At least, putting the rotation angle parameter back if users demand it would be strictly additive to the model so would not be a very disruptive change. The main thing I worry is that users may want to globally rotate the direction by 90 degrees, in case the flow direction was painted as "groove stretch" instead of "highlight stretch". Though of course it can always be done at the level of the flow map.

>On the specification of the tangent space: agreed, will amend the PR to specify the relation between 
👍 

>I agree it makes more sense for the artist to draw the direction of the groves, but as you point out, the vector field will express both equivalently. So in the end I think we can conclude it's irrelevant, which means the decision to how best express it will be left to the authoring tool. On the specification side, we then should probably choose the simplest and clearest specification. I think "NDF stretching along T" would be it (even though it means the opposite of what I originally thought it meant 🙃).

Possibly i'm mistaken, but won't artists usually think of the lines they are painting as corresponding to the "brush" direction in brushed metal? So e.g. to get the thing below, wouldn't they want to paint circles around the base of the pan, rather than radial lines? (As if they are applying the brushing themselves to the metal). In that case, they would be specifying what we refer to a B. 

Personally I'd  vote to swap the meaning of T and B, so that T actually means the groove direction (so B is the NDF highlight stretch direction). This is not what glTF did though, for whatever reason. Ideally we need to hear from more people who are familiar with the actual artist tools and workflows for this (maybe your Substance colleagues for example).

image

virtualzavie:
> Possibly i'm mistaken, but won't artists usually think of the lines they are painting as corresponding to the "brush" direction in brushed metal?

Oh yes, I completely agree. But whether we specify the grooves as being T or B won't matter, since it's up to the authoring tool to decide what a brush stroke of the artist will result in. Arguably, an authoring tool that doesn't give much thought into it will likely map our parameter as is, so that would be an argument to indeed align the grooves with T.

I'll try to gather some feedback internally.

portsmouth:
NB, the Arnold tutorial for that pan render is here:

https://help.autodesk.com/view/ARNOL/ENU/?guid=arnold_for_3ds_max_ax_shading_ax_anisotropic_brushed_html

It's interesting that in this case, he didn't rely much (or at all?) on the tangents, instead the circles were generated using a bump map with circular "scratches" in it (which is kind of like authoring the "anisotropic microfacet distribution" with bumps, rather than the actual NDF).  Which is another valid way to get "anisotropic roughness" at a geometrical level, I suppose.

portsmouth:
>On the additional rotation parameter: I am not convinced this value is worth the cost...  suggest to leave this proposed parameter out at least until we have a clear demand for it.

Agreed.

? On the specification of the tangent space: agreed, will amend the PR to specify the relation between $T$ and $T'$

👍


>On the direction of the grooves vs direction of the highlight: your comment makes me realise I confused what the effect of stretching the NDF would be. I will have to amend the PR to fix that. 

👍

> On the specification side, we then should probably choose the simplest and clearest specification. I think "NDF stretching along T" would be it (even though it means the opposite of what I originally thought it meant 🙃).

>Arguably, an authoring tool that doesn't give much thought into it will likely map our parameter as is, so that would be an argument to indeed align the grooves with T.

I think glTF opted to have the NDF stretch along T.

Personally, I think it's more intuitive though if the *grooves* aligned with T, so the highlight (and thus the NDF stretch) should align with B. 

Though I suppose we should go with whatever convention is most consistent with existing flow-map tools (your Substance colleagues would be the best people to ask).

>On the mapping from 
[0,1] to [-1,1]
(https://github.com/AcademySoftwareFoundation/OpenPBR/issues/155#issuecomment-2009614171) ... I was not sure there was a consensus, but I'll update the PR to reflect this.

Seems like there is a consensus (i.e. do same as in glTF).

>Did I forget anything?

Nope. Only issue to decide is the stretch direction convention. Otherwise, proposed fixes to the spec sound good.

portsmouth (commented):
Do we have a consensus on what to define as the highlight stretch direction, T or B? (Where the groove direction is then the opposite).

As noted, I think highlight=B, groove=T is the most sensible. However, glTF I think does the opposite. So what do we do?

virtualzavie:
I went with what I think is the simplest proposal.

Here are the relevant parts:
image
image
image
image
image
image
image
image

index.html
```diff
@@ -394,7 +394,9 @@
...
+In the general case the roughness is anisotropic, that is the NDF is not circularly symmetric but stretched along some direction in the surface plane, producing an elongation of the specular highlight along that direction. This simulates coherent microscale groove geometry due to processes like scratches or brushing. It is assumed that a reference tangent vector field is defined (via **`geometry_tangent`** and **`geometry_coat_tangent`**). The surface tangent vector indicates the direction along which the NDF is stretched, meaning the microscale grooves tend to be aligned with the orthogonal bitangent.
```
portsmouth, line 397: I think it needs to be clarified here that the “surface tangent vector” *is* the reference tangent vector, from the supplied `geometry_tangent`. Previously they differed since the rotation was applied within the model.

portsmouth:
```suggestion
In the general case the roughness is anisotropic, that is the NDF is not circularly symmetric but stretched along some direction in the surface plane, producing an elongation of the specular highlight along that direction. This simulates coherent microscale groove geometry due to processes like scratches or brushing. It is assumed that a reference tangent vector field is defined (via **`geometry_tangent`** and **`geometry_coat_tangent`**). The reference tangent vector indicates the direction along which the NDF is stretched, meaning the microscale grooves tend to be aligned with the orthogonal bitangent.
```

index.html
```diff
@@ -415,9 +417,7 @@
...
-
-- **`specular_rotation`** for $f_\mathrm{dielectric}$ and $f_\mathrm{conductor}$, and **`coat_rotation`** for $f_\mathrm{coat}$, specify direction of the elongation of the specular highlight. The rotation is normalized to $[0, 1]$ representing angles from 0° to 360° counterclockwise about the normal, with 0° being the original tangent vector direction. Note that the appearance of the specular highlight is identical for rotations between $[0, 0.5]$ and $[0.5, 1]$; this allows the preservation of value when converting to/from other models that support directional anisotropy.
+- **`specular_roughness_anisotropy`** for $f_\mathrm{dielectric}$ and $f_\mathrm{conductor}$, and **`coat_roughness_anisotropy`** for $f_\mathrm{coat}$, specify $a \in [0, 1]$ (the degree to which the NDF is stretched in the direction of the local surface tangent). The resulting NDF $\alpha_t$, $\alpha_b$ parameters are then determined by equation [openpbr-anisotropy-formula]. Note that the appearance of the specular highlight is identical for tangent vectors of opposite directions; this also allows the preservation of value when converting to/from other models that support directional anisotropy.
```
portsmouth, line 420:
```suggestion
- **`specular_roughness_anisotropy`** for $f_\mathrm{dielectric}$ and $f_\mathrm{conductor}$, and **`coat_roughness_anisotropy`** for $f_\mathrm{coat}$, specify $a \in [0, 1]$ (the degree to which the NDF is stretched in the direction of the local surface tangent). The resulting NDF $\alpha_t$, $\alpha_b$ parameters are then determined by equation [openpbr-anisotropy-formula]. Note that the appearance of the specular highlight is identical for tangent vectors of opposite directions; this also allows the preservation of value when converting to or from other models that support directional anisotropy.
```

portsmouth (approved):
LGTM!

PR #171, opened 2024-03-22
Subsurface refactor

portsmouth:
This PR refactors the subsurface discussion to be consistent with our (extensive) discussion in Slack about improving the meaning of the parametrization. None of the parameters have changed (names, types or ranges), but the interpretation of them is re-stated to avoid the issues we were seeing with the old text.

In summary:
  - The RGB  extinction is defined simply as the reciprocal of the RGB radii (r = `subsurface_radius * subsurface_radius_scale`). I noted that this blows up at zero radius, so requires some "regularization".
  - The `subsurface_color` A is defined to parametrize the "multi-scatter" albedo, which is defined being very explicit to disambiguate between the reflection due to Fresnel and due to the medium scattering. This then effectively defines the expected ground truth appearance unambiguously, without appealing to any particular approximation.
  - One particular approximation is written down (the SPI / van de Hulst one), more as an example than a suggestion. We note clearly that this ignores the effect of the dielectric IOR (i.e. assumes IOR=1 basically), and point to references for more detailed/accurate formulas. We also note that the approximation used should take into account any particular assumptions about the boundary BSDF that the renderer makes (e.g. if it assumes diffuse entry/exit, that should inform the remapping formula used, if one wants to still match the ground truth appearance).

With this refactor, the existing implementations we have in Arnold (for example) are reasonable implementations.

image


index.html
```diff
@@ -397,6 +398,7 @@
 D_\mathrm{GGX}(m) \propto \left( 1 + \tan^2\theta_m \left(\frac{\cos^2\phi_m}{\alpha_t^2} + \frac{\sin^2\phi_m}{\alpha_b^2}\right) \right)^{-2}
 \end{equation}
 which reduces to the isotropic form when $\alpha_t = \alpha_b = \alpha$.
+Efficient techniques for sampling BSDFs employing the anisotropic GGX microfacet model are presented in [#Heitz2014a].
```
peterkutz, line 401: This is a nice addition. However the cited paper is no longer the state of the art. I would suggest replacing this citation with these two instead:

`[#Heitz2018]: Eric Heitz. *Sampling the GGX Distribution of Visible Normals*, Journal of Computer Graphics Techniques (2018).`

`[#Dupuy2023]: Jonathan Dupuy, Anis Benyoub. *Sampling Visible GGX Normals with Spherical Caps*, High-Performance Graphics (2023).`

jstone-lucasfilm: Good point, @peterkutz, and indeed it's the Depuy/Benyoub paper that we currently use in MaterialX GLSL/MSL.

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl#L64

index.html
```diff
@@ -594,52 +596,57 @@
 S_\mathrm{subsurface} = \mathrm{Slab}(f_\mathrm{dielectric}, V^\infty_\mathrm{subsurface}) \ .
```
peterkutz, line 596: Minor tangential comment: A few lines above here, on line 593, we say "where in general the exiting ray leaves at a different surface location than the incident ray". It feels like we could probably remove the "in general" part, since mathematically there is a zero probability of exiting at the exact point where we entered. (The probability is non-zero in practice due to rounding, but that seems too academic and negligible to mention.)

index.html
```diff
@@ -594,52 +596,57 @@
...
 \end{equation}
 
-As for the [glossy-diffuse](index.html#model/dielectricbase/glossy-diffuse) slab and the [translucent-base](index.html#model/dielectricbase/translucentbase), the subsurface is bounded by a dielectric interface with BSDF $f_\mathrm{dielectric}$, which generates the primary specular lobe parametrized via the "specular" parameters as described in the Dielectric base section.
+As for the [glossy-diffuse](index.html#model/dielectricbase/glossy-diffuse) slab and the [translucent-base](index.html#model/dielectricbase/translucentbase), the subsurface is bounded by a dielectric interface with BSDF $f_\mathrm{dielectric}$, which generates the primary specular lobe parametrized via the "specular" parameters as described in the Dielectric base section. Superimposed on this is the reflection generated by scattering in the subsurface medium. In this case the subsurface medium $V^\infty_\mathrm{subsurface}$ is given a parametrization which is particularly convenient for controlling the volumetric effect of dense subsurface scattering:
```
peterkutz, line 599: "As for" is a little ambiguous here because it could also be interpreted as "Regarding" (i.e. "Regarding the glossy–diffuse slab and the translucent-base, ....").

But in this case "As for" actually means "As is the case for" or "As in the case of" or "As is true for". So maybe we could tweak the wording to clarify.

peterkutz: I wonder if there's a better word than "Superimposed" here. "Superimposed on this" seems to suggest that the subsurface scattering is on top of the specular reflection. A few alternatives: "Coupled with this" or "Combined with this" or "Alongside this".

index.html
```diff
@@ -594,52 +596,57 @@
...
 
-Superimposed on this is the reflection generated by scattering in the subsurface medium. In this case the subsurface medium $V^\infty_\mathrm{subsurface}$ is given a parametrization which is particularly convenient for controlling the volumetric effect of dense subsurface scattering:
+   - **`subsurface_radius`** * **`subsurface_radius_scale`**: the _mean free path_ (MFP) per RGB channel, $\mathbf{r}$, i.e. the typical distance that light travels in the medium before absorption or scattering. This thus controls the apparent density of the medium. In the limit of zero MFP, the medium tends towards infinite density, and approaches the look of an opaque diffuse surface. Being a length, **`subsurface_radius`** can be any value greater than or equal to zero. For convenience, we make the soft range $[0, 1]$, thus covering common cases where the MFP is typically much lower than the scene length units. The **`subsurface_radius_scale`** controls the color channel dependence of the MFP, and thus this color is visible in the light transmitted through thinner regions of the subsurface volume.
```
peterkutz, line 601: > the typical distance that light travels in the medium before absorption or scattering.

"average" would be a little more precise than "typical" (though "typical" sounds fine if we're intentionally trying to sound more colloquial).

Also, it might potentially be useful to specify that this is the distance _along each individual ray_, maybe by saying something about a ray or a straight line.

Here's a possible alternative wording:

"the average distance that a ray of light travels through the medium before being absorbed or scattered"

peterkutz: > we make the soft range $[0, 1]$, thus covering common cases where the MFP is typically much lower than the scene length units.

Makes sense, although I'm not sure about the assumption that it's typical and common for the MFP to be much lower than the scene length units. It really depends on what those units are and what's being rendered.

peterkutz: > The **`subsurface_radius_scale`** controls the color channel dependence of the MFP, and thus this color is visible in the light transmitted through thinner regions of the subsurface volume.

This color will also be visible in shadowed regions even on an object without thin geometric features. For completeness, I would propose to mention both of these common cases where this color becomes visible. Here's one possible wording:

"and thus this color is visible in light transmitted through thinner features or into shadowed regions of the subsurface volume."

index.html
```diff
@@ -594,52 +596,57 @@
...
+   - **`subsurface_color`**: the observed RGB reflection albedo color taking into account all orders of multiple scattering, $\mathbf{C}$ (where the sense in which this parametrizes the observed color is discussed in detail below).
+   - **`subsurface_anisotropy`**: the medium's phase function, parametrized by the scalar anisotropy $g \in [-1, 1]$.
```
peterkutz, line 603: Do we intentionally want to support $g \in [-1, 1]$ or just $g \in (-1, 1)$? I suppose the implementation is responsible for either clamping the input $g$ or having a special case for perfect forward and backward scattering.

index.html
```diff
@@ -594,52 +596,57 @@
...
 
-   - **`subsurface_radius`** * **`subsurface_radius_scale`**: the desired size of the "blur" on the surface per RGB channel, i.e. roughly the average distance that light travels _along the surface_ per-channel, $\mathbf{r}$, also termed the _diffusion radius_. Being a length, **`subsurface_radius`** can be any value greater than or equal to zero. For convenience, we make the soft range $[0, 1]$, thus covering common cases.
-   - **`subsurface_color`**: the observed RGB reflection albedo color taking into account all orders of multiple scattering, $\mathbf{A}$, also termed the _multi-scatter albedo_.
-   - **`subsurface_anisotropy`**: the medium's phase function (assumed to be the standard Henyey-Greenstein form), parametrized by the scalar anisotropy $g \in [-1, 1]$.
+We now define in detail how these subsurface parameters collectively determine the underlying volumetric parameters (per-channel extinction $\boldsymbol{\mu}_t$ and _single-scattering albedo_ $\boldsymbol{\alpha}$, and scalar phase function anisotropy $g$).
```
peterkutz, line 605: Was it intentional to italicize "single-scattering albedo" but not "extinction"? ("extinction" seems to be an equally technical term and this seems to be the first time it's mentioned.)

index.html
```diff
@@ -594,52 +596,57 @@
...
+This can be broken into two components:
 \begin{equation}
-S(\mathbf{A}) = 4.012 - 15.21 \mathbf{A} + 32.34 \mathbf{A}^2 - 34.68 \mathbf{A}^3 + 13.91 \mathbf{A}^4 \ .
+\mathbf{E}_\mathrm{subsurface} = \mathbf{E}_\mathrm{spec} + \mathbf{E}_\mathrm{multi-scatter} \ .
 \end{equation}
-Similarly the RGB single-scattering albedo $\boldsymbol{\alpha}$ is approximately given by
-\begin{equation}
-\boldsymbol{\alpha}(\mathbf{A}) = 1 - e^{-11.43 \mathbf{A} + 15.38 \mathbf{A}^2 - 13.91 \mathbf{A}^3} \ .
+Here $\mathbf{E}_\mathrm{spec}$ is the normal-direction reflectance of all energy reflected from the dielectric interface _without_ transmission. While $\mathbf{E}_\mathrm{multi-scatter}$ (the _multiple-scattering albedo_) is the normal-direction reflectance of all remaining reflected energy due to transmission through the interface, multiple scattering in the subsurface medium, and transmission back out.
```
peterkutz, line 614: > all energy reflected from the dielectric interface _without_ transmission

Technically it's possible for light to enter and exit the medium during microfacet multiple scattering. However the distances traveled inside the medium in those cases are infinitesimal, so it seems correct for us to assume that light that reflects from the interface after this type of microfacet multiple scattering is still considered part of $\mathbf{E}_\mathrm{spec}$.

index.html
```diff
@@ -594,52 +596,57 @@
...
+We then _define_ the **`subsurface_color`** $\mathbf{C}$ to parametrize the fraction of the energy transmitted through the interface that is transmitted back out and observed, i.e.:
+\begin{equation} \label{subsurface_albedo_constraint}
+\mathbf{E}_\mathrm{multi-scatter} = \left( 1 - \mathbf{E}_\mathrm{spec} \right) \mathbf{C} \ .
 \end{equation}
-The physical RGB scattering coefficient is then given by $\boldsymbol{\mu}_s = \boldsymbol{\alpha}\,\boldsymbol{\mu}_t$.
+Thus since $0 \le \mathbf{C} \le 1$, $\mathbf{E}_\mathrm{subsurface} \le 1$ so energy is always conserved, and if $\mathbf{C}=1$ then $\mathbf{E}_\mathrm{subsurface}=1$ which guarantees that a white **`subsurface_color`** passes the furnace test. According to this definition, the **`subsurface_color`** $\mathbf{C}$ is the observed reflection color (viewed at normal incidence under uniform illumination) in areas where the Fresnel reflection is negligible, and otherwise the observed color is a blend of $\mathbf{C}$ with the gray Fresnel reflection that conserves total reflected energy.
```
peterkutz, line 619: > (viewed at normal incidence under uniform illumination)

We should also mention the semi-infinite homogeneous slab assumption. (Without it, the albedo remapping would have to take the geometry into account, and then this varying remapping would cancel out some of the interesting effects that subsurface scattering produces, such as thin regions looking less saturated than thick regions.)

index.html
```diff
@@ -594,52 +596,57 @@
...
+Given the required multiple-scattering albedo $\mathbf{E}_\mathrm{multi-scatter}$ according to the formula above, then in principle, the single-scattering albedo $\boldsymbol{\alpha}$ which generates the required $\mathbf{E}_\mathrm{multi-scatter}$ can be determined. We do not require any particular theoretical formula for $\boldsymbol{\alpha}(\mathbf{C})$ be used, but the closer it is to satisfying equation [subsurface_albedo_constraint] the closer it is to the ground truth. Ideally, this formula should take into account all the properties of the medium as well as the dielectric interface. A number of such approximate formulas have been derived in the literature. A well-known approximation is due to van de Hulst ([#Kulla2017], [#d'Eon2021]), which assumes an index-matched boundary (i.e. $\mathbf{E}_\mathrm{spec} = 0$, thus $\mathbf{E}_\mathrm{multi-scatter} = \mathbf{C}$), where:
```
peterkutz, line 621: > all the properties of the medium as well as the dielectric interface

I think this part could potentially be misinterpreted to include the shape of the object, however clarifying the semi-infinite homogeneous slab assumption mentioned above could resolve this.

index.html
```diff
@@ -594,52 +596,57 @@
...
 \begin{equation}
-\bar{\boldsymbol{\alpha}} = \frac{\boldsymbol{\alpha}} {1  - g (1 - \boldsymbol{\alpha})} \ ,
+\mathbf{C} = \frac{(1 - \mathbf{s})(1 - 0.139 \mathbf{s})}{1 + 1.17 \mathbf{s}} \ , \quad
+\mathbf{s} = \sqrt{\frac{1 - \boldsymbol{\alpha}}{1 - \boldsymbol{\alpha}\, g}} \ .
 \end{equation}
-and remapped extinction coefficient $\bar{\boldsymbol{\mu}}_t$ by
+Inversion of this gives the per-channel single-scatter albedo $\boldsymbol{\alpha}$ in terms of the **`subsurface_color`** $\mathbf{C}$ and anisotropy $g$:
 \begin{equation}
-\bar{\boldsymbol{\mu}}_t = \frac{1  - g (1 - \boldsymbol{\alpha})}{1 - g} \boldsymbol{\mu}_t \ .
+\boldsymbol{\alpha} = \frac{1 - \mathbf{s}^2}{1 - g\,\mathbf{s}^2} \ , \quad
```
peterkutz, line 628: If I remember correctly, the inclusion of *g* implies in the formula implies that the albedo remapping should compensate for the anisotropy so that the observed color is equal to the specified **`subsurface_color`** regardless of the value of **`subsurface_anisotropy`**. However the extinction should not be affected by the **`subsurface_anisotropy`** (making the volume appear more translucent when it's more forward-scattering).

Is that correct? And do you think it's clear enough or would it be useful to state it more directly in the spec?

index.html
```diff
@@ -594,52 +596,57 @@
...
+\mathbf{s}          = 4.09712 + 4.20863\mathbf{C} - \sqrt{9.59217 + 41.6808C + 17.7126\mathbf{C}^2} \ .
 \end{equation}
-Note that due to the singularity at $g=1$, in practice the anisotropy needs to be limited to be not too close to 1.
+This gives one approximate $\boldsymbol{\alpha}(\mathbf{C})$ mapping, but as noted it ignores the presence of the dielectric boundary. Other approximate forms which take into account the dielectric interface are presented in various sources (e.g. [#d'Eon2021], [#Burley2018]).
+We also note that the relevant approximation may depend on the specific assumptions made about the subsurface interface, for example in some cases renderers may approximate the boundary BSDF via a diffuse lobe, for efficiency and improved convergence. In that case, it would be more appropriate to use an $\boldsymbol{\alpha}(\mathbf{C})$ designed for this specific interface BSDF, in order to satisfy equation [subsurface_albedo_constraint].
+
+The phase function anisotropy $g$ of the medium is taken to be given directly by the **`subsurface_anisotropy`** parameter [^anisotropy_g]. It is assumed that the phase function has the standard Henyey-Greenstein form.
```
peterkutz, line 634: > Henyey-Greenstein

Minor: Many style guides would recommend using en dashes instead of hyphens ("–" instead of "-") for relationships and connections like this one. However it seems fine to use hyphens as long as we're consistent about it.

There's more information about this here if anyone is interested: 
https://en.wikipedia.org/wiki/Dash#Relationships_and_connections

index.html
```diff
@@ -594,52 +596,57 @@
...
 
-Implementations may choose to compute light transport in the subsurface medium either via a fast, approximate bidirectional scattering surface reflectance distribution (BSSRDF) model using e.g. the dipole approximation [#Jensen2001], or more accurately via volumetric path tracing ([#Novak2018], [#Pharr2023]).
+Once the underlying volumetric medium properties $\boldsymbol{\mu}_t$, $\boldsymbol{\alpha}$ and $g$ are determined, implementations may choose to compute light transport in the subsurface medium either via a fast, approximate bidirectional scattering surface reflectance distribution (BSSRDF) model using e.g. the dipole approximation [#Jensen2001], or more accurately via volumetric path tracing ([#Novak2018], [#Pharr2023]).
```
peterkutz, line 636: > fast, approximate BSSRDF model

This seems fine, and approximations seem useful especially in a rasterization context, although I would personally try to avoid making these approximations sound too good for two main reasons:
- The inherent inaccuracies that they involve can result in visual artifacts and extra complexity to cover up those artifacts.
- These approximations are not always faster, especially when efficient sampling is used in the path-traced version or when the path lengths are short because the albedo is low.

Specifically, I'm wondering whether we should remove the word "fast" from the description.

peterkutz: > implementations may choose to compute light transport in the subsurface medium either via a ... BSSRDF model ... or ... via volumetric path tracing

There may be other less common algorithms or variations that could also be used, so I wonder whether we should add the phrase "for example" or something like that. (I'm thinking about other options like photon mapping, VPLs, photon beams, screen-space blurring, etc.)

index.html
```diff
@@ -594,52 +596,57 @@
...
 **`subsurface_weight`**       | Weight       | `float`   | $ [0, 1]      $ |            | $ 0                $ | Mix weight between subsurface and diffuse slabs
-**`subsurface_color`**        | Color        | `color3`  | $ [0, 1]      $ |            | $ (0.8, 0.8, 0.8)  $ | Subsurface color, i.e. the multi-scatter albedo of $V^\infty_\mathrm{subsurface}$
-**`subsurface_radius`**       | Radius       | `float`   | $ [0, \infty) $ | $ [0, 1] $ | $ 1                $ | Length scale of diffusion along the surface
-**`subsurface_radius_scale`** | Radius scale | `vector3` | $ [0, 1]^3    $ |            | $ (1.0, 0.5, 0.25) $ | RGB multiplier to **`subsurface_radius`**, giving the per-channel diffusion length
+**`subsurface_color`**        | Color        | `color3`  | $ [0, 1]      $ |            | $ (0.8, 0.8, 0.8)  $ | Subsurface color, controls the observed reflection color of $V^\infty_\mathrm{subsurface}$
+**`subsurface_radius`**       | Radius       | `float`   | $ [0, \infty) $ | $ [0, 1] $ | $ 1                $ | Length scale of MFP
```
peterkutz, line 648: > Length scale of MFP

Using the word "scale" in the description of the parameter that doesn't include "scale" in the name seems a bit confusing. Maybe we could reword it. Here are some ideas:

"Length of MFP"
"Extinction MFP"
"Scale MFP"
"Base MFP"
"Average distance light travels along a ray, specifically the extinction MFP"

index.html
```diff
@@ -594,52 +596,57 @@
...
 ------------------------------|--------------|-----------|:---------------:|:----------:|:--------------------:|----------------------------------------------
 **`subsurface_weight`**       | Weight       | `float`   | $ [0, 1]      $ |            | $ 0                $ | Mix weight between subsurface and diffuse slabs
-**`subsurface_color`**        | Color        | `color3`  | $ [0, 1]      $ |            | $ (0.8, 0.8, 0.8)  $ | Subsurface color, i.e. the multi-scatter albedo of $V^\infty_\mathrm{subsurface}$
-**`subsurface_radius`**       | Radius       | `float`   | $ [0, \infty) $ | $ [0, 1] $ | $ 1                $ | Length scale of diffusion along the surface
-**`subsurface_radius_scale`** | Radius scale | `vector3` | $ [0, 1]^3    $ |            | $ (1.0, 0.5, 0.25) $ | RGB multiplier to **`subsurface_radius`**, giving the per-channel diffusion length
+**`subsurface_color`**        | Color        | `color3`  | $ [0, 1]      $ |            | $ (0.8, 0.8, 0.8)  $ | Subsurface color, controls the observed reflection color of $V^\infty_\mathrm{subsurface}$
```
peterkutz, line 647: > Subsurface color, controls the observed reflection color

- While "Subsurface color" is an accurate description, it doesn't seem like the most helpful description since the parameter itself is called **`subsurface_color`**.
- It seems like the comma here should technically be a semicolon since it's separating two different descriptions of the same parameter (although if we're using this style consistently it seems fine).

We could consider simply rewording or streamlining this to avoid the redundancy and the need for a semicolon. Here's a proposal for this description:

"The observed reflection color of $V^\infty_\mathrm{subsurface}$"

index.html
```diff
@@ -594,52 +596,57 @@
...
+**`subsurface_radius`**       | Radius       | `float`   | $ [0, \infty) $ | $ [0, 1] $ | $ 1                $ | Length scale of MFP
+**`subsurface_radius_scale`** | Radius scale | `vector3` | $ [0, 1]^3    $ |            | $ (1.0, 0.5, 0.25) $ | RGB multiplier to **`subsurface_radius`**, giving the per-channel MFPs
```
peterkutz, line 649: > (1.0, 0.5, 0.25)

Tangential: Just making sure we're going ahead with this default value as we no longer have any color inversion to compensate for.

(Incidentally, I'm going to miss the "Rayleigh scattering" parameter of the ASM, which let you achieve this type of appearance in a physically based way with a simple slide, without requiring a color to be manually entered. I suppose implementations could provide higher-level controls or presets if they want to use that concept.)

index.html
```diff
@@ -682,11 +689,7 @@
 && \quad\quad \boldsymbol{\mu}_a \leftarrow \boldsymbol{\mu}_a - \mathrm{min}(\boldsymbol{\mu}_a) \ .
 \end{eqnarray}
 After this shifting, the final extinction coefficient is given by the new $\boldsymbol{\mu}_a + \boldsymbol{\mu}_s$.
-This formulation produces volumetric parameters which reproduce reasonably well the independently specified colors of the transmitted and single-scattered light. Finally the medium phase function anisotropy $g \in [-1, 1]$ is given by **`transmission_scatter_anisotropy`** (the standard Henyey-Greenstein phase function is assumed).
-
-!!! Note: Future work
-   The Henyey-Greenstein phase function is not ideal for representing media with both forward and back scattering lobes (as in the Mie phase function). A richer model such as [#Jendersie2023] could be considered in a future version.
-
+This formulation produces volumetric parameters which reproduce reasonably well the independently specified colors of the transmitted and single-scattered light. Finally the medium phase function anisotropy $g \in [-1, 1]$ is given by **`transmission_scatter_anisotropy`** (the standard Henyey-Greenstein phase function is assumed [^anisotropy_g]).
```
peterkutz, line 692: Tangential: On the Adobe side we still don't find the interior volume parameterization particularly intuitive, although I'm not sure if there's time to address it for this first version, and I know there are various tradeoffs. It will be interesting to hear how users feel.

index.html
```diff
@@ -1361,6 +1360,8 @@
 
 [^normalization]:  Omitting normalization factors.
 
+[^anisotropy_g]: Technically, $g$ is the _mean-cosine of deflection_ of the phase-function, which is not specific to the Henyey-Greenstein phase function model [#d'Eon2021].
```
peterkutz, line 1363: Just checking why you decided to put this in a footnote instead of mentioning the "mean cosine" in the main text. Is it because the anisotropy is discussed in two different sections (for subsurface and interior volumes)?

peterkutz: > mean-cosine

> phase-function

These shouldn't be hyphenated. 

("mean" is simply an adjective describing "cosine". And "phase function" would only be hyphenated if being used as a compound adjective, however we decided to not hyphenate even compound adjectives if they are terms of art.)

peterkutz:
@portsmouth This looks great overall. It accurately conveys the conclusions of our last round of discussions. It also makes a number of useful clarifications.

I left a number of comments about specific concerns and suggestions for further improvements.

index.html
portsmouth: Added refences to both in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/edc8216c5e6e672763f4e930cf09ae56358cd130.

index.html
portsmouth: Agreed, done in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/ffcd870d3e8c69a33aa5a632527ce144f42fc048

index.html
portsmouth: In https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/ff04b4249f510c023c8a0573cec9f2dda032540f:

>But in this case "As for" actually means "As is the case for" or "As in the case of" or "As is true for". So maybe we could tweak the wording to clarify.

Tweaked to "As in the cases of"..

> A few alternatives: "Coupled with this" or "Combined with this" or "Alongside this".

I tried something a bit more colloquial (arguably a bit wordy, but clear I hope):

>Combined with this is the reflection generated by light which is transmitted through the dielectric interface into the underlying embedded subsurface medium, where it scatters around and eventually transmits back out.

index.html
portsmouth: >"the average distance that a ray of light travels through the medium before being absorbed or scattered"

Done in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/b1a88da5afcc24dbf5ce06aa3cca54679d223d2a

portsmouth: >Makes sense, although I'm not sure about the assumption that it's typical and common for the MFP to be much lower than the scene length units. It really depends on what those units are and what's being rendered.

True we don't really know of course. Maybe we can soften the language and just say the MFP is just "typically lower" than the scene length units, not "much lower". I think that's probably true, e.g. most of the time SSS is used for characters being modelled in cm or m, and the MFP is of order mm.

index.html
portsmouth: I think in practice one needs to clamp, as various formulas blow up at the limits (and also the rendering gets pathological well before reaching the bounds). It might be convenient though from an artist standpoint to be able to dial all the way to the bounds, and still get something sensible, relying on an internal clamp. It seems like a good idea to explicitly mention that implementations may internally clamp the anisotropy to avoid such pathologies.

index.html
portsmouth: Good spot. Fixed in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/03ff0b551b6a2a21b1536924b4217366006b2f88

index.html
portsmouth: I had in mind actually that even paths at the microsurface level that enter the medium should count as part of the multi-scatter albedo. So that the $\mathbf{E}_\mathrm{spec}$ term is rigorously composed only of paths which reflect (once or multiple times) on the microsurface but never transmit (so never touch the medium, even at the microsurface level). 

But thinking about it now, at the microsurface scale it seems appropriate to assume that the medium MFP is effectively infinite, compared to the length scale of the distances between surface scatterings (like you said). So at each macroscopic vertex in the path-trace, scattering on the microsurface generates reflection and transmission lobes, corresponding to paths which reflect or transmit to "infinity" (on the microscale) after the interaction, effectively oblivious to any volumetric media. Then the volumetric scatterings and subsequent surface bounces should happen at a larger macroscopic scale.

So I think it's more consistent then to say that the $\mathbf{E}_\mathrm{spec}$ term represents the reflection lobe of the initial macroscopic bounce  -- so it should include both single and multiple-scattering on the microsurface. The relevant distinction between $\mathbf{E}_\mathrm{spec}$ and $\mathbf{E}_\mathrm{multi-scatter}$ is just that the latter lobe consists of paths which left the initial micro-surface interaction into the lower hemisphere, so then are able to subsequently scatter in the medium (though may not, e.g. if the medium density is zero). 

This seems a tad over-academic. I'm not sure it would make much of a difference to how implementations should derive their mappings (assuming they even attempt to, rather than using one from the literature off-the-shelf such as the stated one, which is more likely). So I'll attempt to word it in a way that unambiguously describes what I wrote above, but without being overly pedantic.

index.html
peterkutz: Makes sense. I think softening the language would be sufficient.

> thus covering common cases where the MFP is typically much lower than the scene length units

It's mostly the combination of "common", "typically", and "much" that makes me skeptical of the statement. "typically" seems redundant with "common", and "much lower" makes me question why the soft max distance even needs to be as high as 1. How about this instead:

"thus covering common cases where the MFP is lower than the scene length units"

index.html
peterkutz: That makes sense.

> It seems like a good idea to explicitly mention that implementations may internally clamp the anisotropy to avoid such pathologies.

That would be nice (assuming it's not obvious).

Specular reflection has similar issues when the roughness approaches zero. Not sure if we need to mention that or if it's implied that implementations will have to handle certain mathematical and numerical corner cases.

index.html
peterkutz: Yes, exactly, I agree with all of that reasoning.

When using approximations to microfacet multiple scattering, it's not even known what sequence of reflection and transmission events occurred at the microsurface level to produce the overall multiple-scattering reflection and transmission lobes. And when using a full random walk, the distances between microsurface events are infinitesimal anyway. So in practice it shouldn't really make any difference whether $\mathbf{E}_\mathrm{spec}$ contains secondary microfacet scattering or not.

> So I'll attempt to word it in a way that unambiguously describes what I wrote above, but without being overly pedantic.

Sounds good. A full discussion seems too academic, but since the description is pretty rigorous and technical already, it seems worthwhile to at least add a few words to clarify.

Even just adding the word "macroscopic" could potentially solve the problem (which is done like this one other place in the spec already):

"all energy reflected from the dielectric interface without (macroscopic) transmission"

index.html
portsmouth: In https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/3b0e2f1fddf6b017a893adb7b156a223e5e85ef3 altered it to:

>For convenience, we make the soft range [0,1], thus covering common cases such as skin where the MFP is lower than the scene length units.

I think it helps to give the informal example of "skin", as it seems intuitive that the MFP of skin is low.

index.html
portsmouth: Suggested clarification:
```suggestion
   - **`subsurface_anisotropy`**: the medium's phase function, parametrized by the scalar anisotropy $g \in [-1, 1]$. In practice implementations may want to internally clamp the anisotropy as the phase function becomes ill-defined and the rendering unstable near the -1, 1 limits.
```

peterkutz: That clarification sounds good to me.

index.html
peterkutz: Sounds good.

index.html
portsmouth: >Even just adding the word "macroscopic" could potentially solve the problem (which is done like this one other place in the spec already):

Fixed along those lines in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/47edc8cdcd80d269c946c610fe1b016bd44cf446

index.html
portsmouth: Yep good point, that does need to be made explicit. Arguably this is implicit in the definition of the Slab concept, which we introduced saying:

>Each slab consists of a homogeneous dielectric or conducting medium $V$ with conceptually unbounded extent horizontally and a finite vertical extent ...

But it doesn't hurt to re-emphasize it. This makes the wording sound a bit pedantic and technical, but it's worth stating to be fully correct. 

In https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/59ed616715b90877bdeddbf1f7935fc78956bf56 I made the language more precise as follows:

>We then _define_ the **`subsurface_color`** $\mathbf{C}$ to parametrize, assuming correct physical light transport and infinite homogeneous slab geometry, the fraction of the energy transmitted through the interface that is transmitted back out and observed ...

Probably we should add similar clarification in the glossy-diffuse section as well.

index.html
portsmouth: Hopefully the re-wording mentioned in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171#discussion_r1571165046 addresses this.

index.html
portsmouth: Yes $g$ should be taken into account in the albedo remapping, since we said:

>Given the required multiple-scattering albedo ... according to the formula above, then in principle, the single-scattering albedo ... which generates the required [multiple-scattering albedo] can be determined...  Ideally, this formula should take into account all the properties of the medium as well as the dielectric interface.

So whatever the anisotropy is, the observed color should match the input. We did say fairly explicitly that the observed color is only affecting the single scattering albedo, but I think it could be further clarified if in this sentence

>Given the required multiple-scattering albedo $\mathbf{E}_\mathrm{multi-scatter}$   according to the formula above, 

>then in principle, the single-scattering albedo $\boldsymbol{\alpha}$ which generates the required $\mathbf{E}_\mathrm{multi-scatter}$ can be determined 

we add the the clarification:

>(assuming the given extinction $\boldsymbol{\mu}_t$ and anisotropy $g$).

I pushed that in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/44cb7a2a88cbb634c745d100c47720555d76b832.

index.html
portsmouth: In https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/commits/9311dfd99ee1133cf3f2791e5521c53754ffe425 I fixed up the hyphens to en dashes, for Henyey-Greenstein. We can make similar fixes elsewhere as needed.

index.html
portsmouth: Suggested rewording:

>Once the underlying volumetric medium properties $\boldsymbol{\mu}_t$, $\boldsymbol{\alpha}$ and $g$ are determined, implementations are free to choose how to compute the light transport in the subsurface medium according to their use case, for example accurately via volumetric path tracing ([#Novak2018], [#Pharr2023]), or approximately via a  bidirectional scattering surface reflectance distribution (BSSRDF) model using e.g. the dipole approximation [#Jensen2001].


```suggestion
Once the underlying volumetric medium properties $\boldsymbol{\mu}_t$, $\boldsymbol{\alpha}$ and $g$ are determined, implementations are free to choose how to compute the light transport in the subsurface medium according to their use case, for example accurately via volumetric path tracing ([#Novak2018], [#Pharr2023]), or approximately via a  bidirectional scattering surface reflectance distribution (BSSRDF) model using e.g. the dipole approximation [#Jensen2001].```

File: index.html (hunk 14 above)
Comment (portsmouth) [reply index.html]: Good catch, made your suggested change in 416d54e40e47b247ceaad35c406ca3664ec29fdf

File: index.html (hunk 13 above)
Comment (portsmouth) [reply index.html:644]: This is maybe from habit, as "length scale" is a common phrase in physics (meaning the rough magnitude of the distance at which some effect is relevant).

Perhaps the term "scale" is confusing though since there is a parameter `subsurface_radius_scale` as well. It may make more sense to call the scalar radius the "scale", and the color just "radius" -- which is what we did in Standard Surface (except `subsurface_scale` was the name of the scalar length).

I do think it makes more sense for something called a "scale" to be a length scale (which in the physics terminology, is an actual length).

File: index.html (hunk 15 above)
Comment (portsmouth) [reply index.html:645]: The reasoning for this value is:

>Note that the default value of subsurface_radius_scale is set at (1,0.5,0.25)
 to approximate the effect of Rayleigh scattering. If we roughly approximate the wavelength corresponding to each of the RGB channels as $\lambda$/nm≈650,550,450 and assume the radii scale like the reciprocal of the extinction, then since in Rayleigh scattering the extinction scales like $\lambda^{-4}$ for light of wavelength $\lambda$ the expected relative magnitudes of the radii are $(1,(550/650)^4, (450/650)^4) \approx (1,0.5,0.25)$, hence the chosen default. This provides a slightly more realistic default look for the subsurface than resulting from gray radii.

Does it sound plausible to you? I noticed that the saturation of the resulting color does vary quite a bit depending on how dense the subsurface medium is.

Comment (portsmouth) [reply index.html:645]: >(Incidentally, I'm going to miss the "Rayleigh scattering" parameter of the ASM, which let you achieve this type of appearance in a physically based way with a simple slider, without requiring a color to be manually entered. I suppose implementations could provide higher-level controls or presets if they want to use that concept.)

Yeah, it would be good to know what the logic for this higher level node should be to replicate the ASM control. We could then provide it e.g. in MaterialX.

File: index.html (hunk 16 above)
Comment (portsmouth) [reply index.html]: Sure, of course I'm also uncomfortable having these two different parametrizations in the same model.

That said, it does make some logical sense that the parametrization should differ for the case of very dense and very thin media. The albedo remapping at least cannot be expected to work for very thin translucent volumes, so it seems inappropriate to use that when authoring e.g. colored glass.

(This could have been handled within the framework of a unified volume though, just it was too big a step to make that happen this time round).

File: index.html (hunk 17 above)
Comment (portsmouth) [reply index.html]: >Just checking why you decided to put this in a footnote instead of mentioning the "mean cosine" in the main text. Is it because the anisotropy is discussed in two different sections (for subsurface and interior volumes)?

Just because this is a technical detail that seems out of place in the main text. It's not a fact you even to know to implement the Henyey-Greenstein model.

Comment (portsmouth) [reply index.html]: >These shouldn't be hyphenated.

Fixed in fc4c5d25705bb7560751b65b18801d44c1c4ef83.

**Comment by portsmouth** [comment]
@peterkutz @fpsunflower Do you mind signing off that these formulas are stated correctly?

image

**Comment by fpsunflower** [comment]
LGTM!

File: index.html (hunk 8 above)
Comment (peterkutz) [reply index.html:632]: Good point that the whole model has a slab assumption. It makes sense to clarify it in this subsurface section anyway since the subsurface light transport controlled by the subsurface parameters happens at the macroscopic scale.

File: index.html (hunk 16 above)
Comment (peterkutz) [reply index.html]: Yes, I agree with all of that.

File: index.html (hunk 17 above)
Comment (peterkutz) [reply index.html]: Makes sense.

**Comment by peterkutz** [comment]
Just to double-check, we want the anisotropy to be incorporated into the color parameterization, since that's a multiple-scattering perceptual parameterization, but we don't want the anisotropy to be incorporated into the scale parameterization, since that's a single-scattering non-perceptual parameterization. Is that right?

File: index.html [hunk 18]
```diff
@@ -608,7 +608,7 @@
...
+Here $\mathbf{E}_\mathrm{spec}$ is the normal-direction reflectance of all energy reflected from the dielectric interface _without_ macroscopic transmission. While the _multiple-scattering albedo_ $\mathbf{E}_\mathrm{multi-scatter}$ is the normal-direction reflectance of all remaining reflected energy due to (macroscopic) transmission through the interface, multiple scattering in the subsurface medium, and transmission back out. We then _define_ the **`subsurface_color`** $\mathbf{C}$ to parametrize, assuming correct physical light transport and infinite homogeneous slab geometry, the fraction of the energy transmitted through the interface that is transmitted back out and observed, i.e.:
```
peterkutz, line 628: Would "semi-infinite" be slightly more clear than "infinite"?

portsmouth:
>Just to double-check, we want the anisotropy to be incorporated into the color parameterization, since that's a multiple-scattering perceptual parameterization, but we don't want the anisotropy to be incorporated into the scale parameterization, since that's a single-scattering non-perceptual parameterization. Is that right?

I think it's not clear what "anisotropy ... incorporated into the scale parameterization" would do exactly, in terms of the desired effect on the appearance. Currently the anisotropy control will just do what it physically does (bias the scattering forwards or backwards), while we also adjust the single-scatter albedo to achieve the desired multi-scatter color. Of course, as we discussed before this means the anisotropy mostly seems to have the effect of making the medium more transmissive (which we previously tried to compensate for), though it seems like a somewhat familiar effect perceptually so good to leave this behavior in.

peterkutz:
Makes sense. I also reviewed your previous reply about this question here:
https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/files#r1571187628

It seems clear that the color mapping should take the anisotropy and dielectric specular interface properties into account. 

> Of course, as we discussed before this means the anisotropy mostly seems to have the effect of making the medium more or less transmissive (which we previously tried to compensate for)

Right. While I don't see it explicitly mentioned, I suppose it's clear enough in the text that the anisotropy will not effect the radius->extinction mapping, which implies that the anisotropy will make the medium appear more or less translucent.

peterkutz (approved):
Thanks for making all of the clarifications and improvements @portsmouth . This looks ready to merge to me!

jstone-lucasfilm (approved):
These changes look good to me!  Are there any corresponding updates required for our reference implementation, or are we ready to merge this work to main?

portsmouth:
> Makes sense. I also reviewed your previous reply about this question here: https://github.com/AcademySoftwareFoundation/OpenPBR/pull/171/files#r1571187628
> 
> It seems clear that the color mapping should take the anisotropy and dielectric specular interface properties into account.
> 
> > Of course, as we discussed before this means the anisotropy mostly seems to have the effect of making the medium more or less transmissive (which we previously tried to compensate for)
> 
> Right. While I don't see it explicitly mentioned, I suppose it's clear enough in the text that the anisotropy will not effect the radius->extinction mapping, which implies that the anisotropy will make the medium appear more or less translucent.

It might be worth adding a sentence to make that explicit, along the lines of:
>Note that the extinction $\boldsymbol{\mu}_t$ is determined only by the scaled subsurface radii $\mathbf{r}$, independent of the anisotropy $g$. Thus the anisotropy has the familiar physical effect of allowing light to transmit more deeply into the medium when positive than when negative.

index.html
```diff
@@ -605,59 +604,63 @@
...
 
-To minimize the entanglement and side effects of these parameters, they should each affect only the corresponding aspect of the subsurface scattering, e.g. the color should not affect the size of the blur and the anisotropy should not affect the size of the blur or the color. We stipulate here a mapping which achieves these requirements and reproduces the specified RGB diffusion profile size (width on the surface) $\mathbf{r}$ and RGB multi-scatter albedo $\mathbf{A}$, by combining the empirically-derived mapping of [#Burley2018] with similarity theory. This defines in detail how the subsurface parameters collectively determine the underlying volumetric parameters (extinction $\boldsymbol{\mu}_t$, single-scattering albedo $\boldsymbol{\alpha}$, and phase function anisotropy $g$).
+The extinction coefficient is simply given by the reciprocal of the MFP per channel: $\boldsymbol{\mu}_t = 1/\mathbf{r}$. (Note, this may need to be regularized in the limit $\mathbf{r} \rightarrow 0$ to avoid numerical issues).
```
portsmouth, line 621:
```suggestion
The extinction coefficient is simply given by the reciprocal of the MFP per channel: $\boldsymbol{\mu}_t = 1/\mathbf{r}$. (Note, this may need to be regularized in the limit $\mathbf{r} \rightarrow 0$ to avoid numerical issues). As the extinction $\boldsymbol{\mu}_t$ is determined only by the scaled subsurface radii $\mathbf{r}$, independent of the anisotropy $g$, the anisotropy has the familiar physical effect of allowing light to transmit more deeply into the medium when positive than when negative.
```

PR #173, opened 2024-04-16
Remove unused transmission roughness from reference graph

portsmouth:


jstone-lucasfilm:
Isn't the `transmission_roughness` value still referenced in the `dielectric_transmission` node of the graph?

https://github.com/AcademySoftwareFoundation/OpenPBR/blob/main/reference/open_pbr_surface.mtlx#L412

virtualzavie (commented):
I have the same comment as Jonathan.

portsmouth:
OK yup, just a misunderstanding on my part. Indeed, a rough coat should roughen the transmission lobe. This needs to be mentioned as part of https://github.com/AcademySoftwareFoundation/OpenPBR/pull/172,  i.e. exactly what parameters are modified by the coat roughening effect.

portsmouth:
OK yup, just a misunderstanding on my part. Indeed, a rough coat should roughen the transmission lobe. This needs to be mentioned as part of https://github.com/AcademySoftwareFoundation/OpenPBR/pull/172,  i.e. exactly what in the model is modified by the coat roughening effect.

portsmouth:
Actually, no, it does make sense. The `transmission_roughness` in this can be replaced with `main_roughness`:
```xml
   <dielectric_bsdf name="dielectric_transmission" type="BSDF">
      <input name="weight" type="float" value="1.0" />
      <input name="tint" type="color3" nodename="if_transmission_tint" />
      <input name="ior" type="float" nodename="modulated_eta_s" />
      <input name="roughness" type="vector2" nodename="transmission_roughness" />
      <input name="tangent" type="vector3" nodename="main_tangent" />
      <input name="distribution" type="string" value="ggx" />
      <input name="scatter_mode" type="string" value="T" />
    </dielectric_bsdf>
``` 
As there is no need to distinguish between the "transmission roughness" and the "reflection roughness" of the dielectric (in OpenPBR).

jstone-lucasfilm (approved):

PR #174, opened 2024-04-16
Define Oren-Nayar lobe explicitly as EON model (i.e. energy-compensated Fujii)

portsmouth:
As discussed extensively on Slack, our specification of the diffuse lobe is incomplete since we don't make it clear what form of the Oren-Nayar model to take, and there are several options.

As argued on Slack, I think the model proposed by Fujii makes the most sense, as it: 
   - fixes the dark ring artifacts in the original "qualitative Oren-Nayar"
   - is less dark than the original "qualitative Oren-Nayar" (energy preserving at grazing)
   - $\sigma$ parameter has more obvious meaning as the $[0,1]$ "weight of non-Lambertian-ness".
   - looks better in renders (subjectively)

It is also easy to add straightforward energy compensation so that:
   - the model is fully energy preserving, so passes a white furnace test
   - exhibits saturation of the albedo, as physically expected for a multiple scattering microfacet model

This model is also very easy to implement (no tabulation) and efficient (even more-so than qualitative Oren-Nayar), via an accurate fit thanks to the analytical solution for the albedo.

NB, this model is fairly obvious, but apparently novel, so could be publishable. (Ideally, should be as a companion to the spec). 

image

image

image


portsmouth:
Some example renders showing results with:

   - Lambert model (all other models collapse to this at zero roughness)
   - Original "qualitative" Oren-Nayar (QON), max roughness.
   - Fujii's proposed modification (FON), max roughness.
   - Fujii's model with analytical energy compensation (EON), max roughness.

| Lambert      | QON            | FON             | EON
| ------------- | ------------- | ------------- | ------------- |
| image| image | image | image |PBR/assets/2774364/18ac3b73-6e2f-4bc1-8330-6d28765a61a9) |

| Lambert      | QON            | FON             | EON
| ------------- | ------------- | ------------- | ------------- |
| image | image | image | image |
 
| Lambert      | QON            | FON             | EON
| ------------- | ------------- | ------------- | ------------- |
| image | image | image | image |

jstone-lucasfilm:
Although I love the look of the energy-compensated Fujii model, there are some practical motivations for considering this a future update to OpenPBR, rather than a feature of the 1.0 version that is scheduled for release at FMX/SIGGRAPH.

One of these motivations is the current implementations of Oren-Nayar in popular closure-based shading languages such as OSL and MDL, which matches the non-energy-compensated QON model.  If it were possible to implement your new energy-compensated Fujii model as a pure MaterialX graph, with the existing Oren-Nayar closures used under the hood, then I would have no objection to including this new feature in OpenPBR.  But any changes that require rewrites of closures across the industry should be discussed and agreed upon in an industry-wide forum before their inclusion in OpenPBR.

portsmouth:
>If it were possible to implement your new energy-compensated Fujii model as a pure MaterialX graph, with the existing Oren-Nayar closures used under the hood...

I note that MaterialX doesn't itself define what Oren-Nayar model it uses (i.e. there are no formulas or references), there is just some code in OSL testrender and MDL. On what basis was the formula in that code chosen (and what model is it)? 

We do need to state some concrete form of the Oren-Nayar model. I strongly suspect that whatever model we choose now, will persist for the lifetime of OpenPBR. (Arguing to change it once it is used in the wild, will be exponentially harder).

Forcing OpenPBR to match the current state of the OSL testrender code, just as a short term convenience rather than based on the merits of the model, seems wrong to me.

jstone-lucasfilm:
Alas, it's both of the closure-based shading languages in our industry (OSL and MDL) that harmonize on QON in their open implementations, and MaterialX simply follows their example.  So we'll need to reach consensus with both the OSL and MDL teams before proposing a change.

I really do like the look of your  energy-compensated Fujii model proposed above, but it seems far too late in the development of OpenPBR (one week before the FMX conference) to propose a change that will require industry agreement.

With our earlier proposals for F82 tint and Zeltner/Burley sheen, we gave the industry more than 6 months to reach consensus and begin authoring their own implementations in preparations for OpenPBR, and I would strongly recommend that we give the industry similar lead time for proposing a new Oren-Nayar model in a future update to OpenPBR.

portsmouth:
>Alas, it's both of the closure-based shading languages in our industry (OSL and MDL) that harmonize on QON in their open implementations, and MaterialX simply follows their example.

But MaterialX spec doesn't say the model is QON, does it? What do you mean by MaterialX follows their example.

jstone-lucasfilm:
As far as I know, neither of the major closure-based shading languages state the definition of Oren-Nayar in their specification, and it's true that MaterialX does not provide its own definition in its specification.

But both OSL and MDL make explicit choices for Oren-Nayar in their open implementations on GitHub, and MaterialX follows their example in its GLSL/ESSL/MSL implementation of Oren-Nayar.

peterkutz:
> If it were possible to implement your new energy-compensated Fujii model as a pure MaterialX graph, with the existing Oren-Nayar closures used under the hood, then I would have no objection to including this new feature in OpenPBR.

Is this a possibility that has been explored?

Specifically, is there any way we could describe the diffuse lobe as a multilple scattering model based on Lambertian microfacets, treat this energy-compensated Fujii model as the ground truth, and also provide an approximation based on QON?

For the approximation, for example, could we for apply a novel scale factor to QON or mix it with a Lambertian lobe that accounts for the missing energy?

virtualzavie (approved):
This looks really good.

As for the closure based languages, as Peter notes, isn't it possible to emulate the model with existing closures?
But even it isn't, choosing an energy conserving model seems the way forward, and like Jamie I'm concerned we will get stuck with an inadequate model otherwise.

jstone-lucasfilm:
@virtualzavie If this math can be accurately modeled using existing closures, then by all means let's make this change to our reference implementation, and we can make it available to the community for validation and visual testing.  If it cannot be accurately modeled in this way, then we should move forward with a proposal for a new closure in OSL and MDL, so that OpenPBR can leverage this closure in its reference graph.

portsmouth:
>As for the closure based languages, as Peter notes, isn't it possible to emulate the model with existing closures?

>@virtualzavie If this math can be accurately modeled using existing closures, then by all means let's make this change to our reference implementation, and we can make it available to the community for validation and visual testing. If it cannot be accurately modeled in this way, then we should move forward with a proposal for a new closure in OSL and MDL, so that OpenPBR can leverage this closure in its reference graph.


No, I don't think it can be accurately modeled using the existing closures. Since the existing MaterialX closure doesn't define what Oren-Nayar model exactly it uses, but in practice in the OSL/GLSL implementations (and I assume MDL) though, it uses what I called the QON model, and there is no way to map from QON to FON by changing the input parameters.
  
If we ignore that and just implement the energy compensation in the graph, then we would have some approximation, but it would be some sort of hybrid that doesn't quite pass the white furnace test, and also will be a huge mess in the XML graph.

It would be much better to propose it as a new closure, probably via a flag on the existing Oren-Nayar closure (can just be a mode selection string I think, i.e. make the current one be "QON", and add "EON").

portsmouth:
Some renders of the (max roughness) EON model implemented in Arnold, compared to the existing max roughness diffuse (QON) and Lambert:

White furnace test, left to right Lambert, QON, EON:
white_furnace

Glossy-diffuse render, left to right Lambert, QON, EON:
glossy_diffuse

Cloth ball renders:

| Lambert    | QON           | EON             |
| :-------------: | :-------------: | :-------------: |
| cloth_lambert |  cloth_QON | cloth_EON |

index.html
```diff
@@ -543,7 +543,8 @@
 
 ### Glossy-diffuse
 
-The glossy-diffuse slab represents a dielectric with rough GGX [microfacet](index.html#model/microfacetmodel) surface BSDF (whose "specular" parameters are described in the Dielectric base section), embedding a semi-infinite bulk of extremely dense scattering material.
+The glossy-diffuse slab represents a dielectric with rough GGX [microfacet](index.html#model/microfacetmodel) surface BSDF (whose "specular" parameters are described in the Dielectric base section), embedding a semi-infinite bulk of extremely dense scattering material. The resulting BRDF of the glossy-diffuse slab is the combination of a "glossy" specular lobe provided by reflection from the dielectric interface, and a diffuse lobe provided by bounces off the substrate. This models for example the reflection from shiny, totally opaque surfaces such as dense plastic, rock, and concrete.
```
virtualzavie, line 546: Minor remark: the Dielectric-base section already describes the dielectric surface, so mabe we don't need to repeat that here, especially since we don't in the Subsurface and Translucent base sub-sections.

The reason I noticed this was because I first understood this paragraph as saying both the dielectric and the diffuse had a GGX NDF, and it took me a couple more reading to understand the importance of "embedding" in this description.

portsmouth: We do mention it also in the Subsurface and Translucent base sections:
>As in the cases of the glossy-diffuse slab and the translucent-base, the subsurface is bounded by a dielectric interface with BSDF fdielectric which generates the primary specular reflection lobe parametrized via the “specular” parameters as described in the Dielectric base section. Combined with this is the reflection generated by light which is transmitted through the dielectric interface into the underlying embedded subsurface medium, where it scatters around and eventually transmits back out.

>As for the dielectric slabs defined in the Glossy-diffuse and Subsurface section, the top interface is described by a rough GGX microfacet BSDF fdielectric whose “specular” parameters are described in the Dielectric base section.

So for consistency I think we should mention it as well here.
I agree the wording could be better though, to make it clear that the "dielectric with rough GGX microfacet surface" being referred to is the same one as in the other sections (and we should link to the Dielectric base section).

virtualzavie: To be fair, we're in nitpicking territory here, and I'm perfectly fine with the PR being merged as it is.

virtualzavie: Thanks for clarifying it.

jstone-lucasfilm:
I'm looking forward to today's MaterialX TSC meeting, where @portsmouth and @peterkutz plan to propose this functionality as a feature of the `oren_nayar_diffuse_bsdf` node in MaterialX and its corresponding closures in shading languages.

Once we have a sense of the future direction of this feature in the industry, we should document this in our OpenPBR specification, so that readers have an understanding of how this feature will be implemented in future versions of MaterialX/OSL/MDL, and giving a sense of what the corresponding changes to our reference implementation will look like.

portsmouth:
@jstone-lucasfilm anything else needed in this PR in order to be mergable, since I think we agreed to go ahead with it following the MaterialX TSC?

jstone-lucasfilm:
@portsmouth I was imagining that we'd post the new energy-compensated diffuse model as an external whitepaper, so that both the MaterialX and OpenPBR specifications can reference it in a consistent way.  This seems like a better precedent for us to follow, aligned with the other novel BSDF lobes that we're adding to both MaterialX and OpenPBR, and it's completely fine if the external whitepaper evolves in the build-up to SIGGRAPH 2024.

portsmouth:
> @portsmouth I was imagining that we'd post the new energy-compensated diffuse model as an external whitepaper, so that both the MaterialX and OpenPBR specifications can reference it in a consistent way. This seems like a better precedent for us to follow, aligned with the other novel BSDF lobes that we're adding to both MaterialX and OpenPBR, and it's completely fine if the external whitepaper evolves in the build-up to SIGGRAPH 2024.

I agree with that, but we are trying to get it published in a journal, so can't reference it immediately, and it would be confusing to reference a half-baked whitepaper in the interim. As noted on Slack, i'd prefer to just move forward with this PR (with the concise summary of the formulas, so then the spec is logically consistent), and then as soon as the paper becomes available, swap out the formulas for a reference to the paper which will provide the formulas.

jstone-lucasfilm (approved):
Thanks @portsmouth and @peterkutz, this looks good to me, and let's move forward with the merge to main.

etheory:
Very cool, I derived this independently but never did anything with it, looks exactly the same, nice: https://www.desmos.com/calculator/extaj6cnzr.

PR #181, opened 2024-05-08
Remove Future work notes and left over comments

virtualzavie:
It was my understanding during the discussions that we prefered to rely on PRs and issues regarding future work, rather than notes in the specification. This PR therefore removes the "Future work" notes from the document, as well as a left over comment (former parameters of the coat).


jstone-lucasfilm (approved):
This looks like a good change to me, thanks @virtualzavie!

PR #185, opened 2024-05-08
Fix IOR schematic figure in thin-film section

portsmouth:
Fixes image link (name was wrong).


portsmouth:
@jstone-lucasfilm would be to merge this ASAP, as the live spec has a broken figure which this fixes.

jstone-lucasfilm (approved):
Looks good to me, thanks @portsmouth!

PR #186, opened 2024-05-10
Remove conclusion section

portsmouth:
Since it's a specification, not a paper making an argument based on methods and results..  Also this "conclusion" is fairly content-free, so not helpful.


jstone-lucasfilm (approved):
Good idea, @portsmouth, and let's move forward with this change.

PR #188, opened 2024-05-11
Upgrade reference implementation to 1.39

jstone-lucasfilm:
This changelist upgrades the reference implementation of OpenPBR to MaterialX 1.39, enabling the following improvements:

- Enable Hoffman/Schlick Fresnel for metals, connecting the new color82 input of generalized_schlick_bsdf.
- Update thin-film logic to include unit conversion and more accurate layering.
- Update swizzle nodes and channel attributes for 1.39 syntax.
- Remove a handful of default-valued inputs for clarity.

### DIFF
--- a/reference/open_pbr_surface.mtlx
+++ b/reference/open_pbr_surface.mtlx
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<materialx version="1.38">
+<materialx version="1.39">
   <!--
     OpenPBR Surface node definition
   -->
@@ -163,7 +163,6 @@
       <input name="in2" type="float" value="0.0" />
     </max>
     <oren_nayar_diffuse_bsdf name="subsurface_thin_walled_reflection_bsdf" type="BSDF">
-      <input name="weight" type="float" value="1.0" />
       <input name="color" type="color3" nodename="subsurface_color_nonnegative" />
       <input name="roughness" type="float" interfacename="base_diffuse_roughness" />
       <input name="normal" type="vector3" interfacename="geometry_normal" />
@@ -181,7 +180,6 @@
       <input name="in2" type="color3" nodename="subsurface_thin_walled_brdf_factor" />
     </multiply>
     <translucent_bsdf name="subsurface_thin_walled_transmission_bsdf" type="BSDF">
-      <input name="weight" type="float" value="1.0" />
       <input name="color" type="color3" nodename="subsurface_color_nonnegative" />
       <input name="normal" type="vector3" interfacename="geometry_normal" />
     </translucent_bsdf>
@@ -212,7 +210,6 @@
       <input name="in2" type="float" interfacename="subsurface_radius" />
     </multiply>
     <subsurface_bsdf name="subsurface_bsdf" type="BSDF">
-      <input name="weight" type="float" value="1.0" />
       <input name="color" type="color3" nodename="subsurface_color_nonnegative" />
       <input name="radius" type="vector3" nodename="subsurface_radius_scaled" />
       <input name="anisotropy" type="float" interfacename="subsurface_anisotropy" />
@@ -324,6 +321,12 @@
       <input name="anisotropy" type="float" interfacename="transmission_scatter_anisotropy" />
     </anisotropic_vdf>
 
+    <!-- Thin-film Thickness -->
+    <multiply name="thin_film_thickness_nm" type="float">
+      <input name="in1" type="float" interfacename="thin_film_thickness" />
+      <input name="in2" type="float" value="1000.0" />
+    </multiply>
+
     <!-- Dielectric Base -->
     <divide name="specular_to_coat_ior_ratio" type="float">
       <input name="in1" type="float" interfacename="specular_ior" />
@@ -388,13 +391,11 @@
       <input name="in2" type="color3" interfacename="transmission_color" />
     </ifgreater>
     <dielectric_bsdf name="dielectric_transmission" type="BSDF">
-      <input name="weight" type="float" value="1.0" />
       <input name="tint" type="color3" nodename="if_transmission_tint" />
       <input name="ior" type="float" nodename="modulated_eta_s" />
       <input name="roughness" type="vector2" nodename="main_roughness" />
       <input name="normal" type="vector3" interfacename="geometry_normal" />
       <input name="tangent" type="vector3" nodename="main_tangent" />
-      <input name="distribution" type="string" value="ggx" />
       <input name="scatter_mode" type="string" value="T" />
     </dielectric_bsdf>
     <layer name="dielectric_volume_transmission" type="BSDF">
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
The latest reference implementation has been validated in the 1.39 build of the MaterialX Viewer, with two adjusted examples shown below:

Brass material with adjusted Specular Color (i.e. F82 edge tint):
OpenPbr_Brass_AdjustedSpecularColor

Soap Bubble material with adjusted Thin-Film Thickness:
OpenPbr_SoapBubble_AdjustedThickness

portsmouth:
```xml
<multiply name="metal_edgecolor" type="color3">
      <input name="in1" type="color3" interfacename="specular_color" />
      <input name="in2" type="float" interfacename="specular_weight" />
</multiply>
```
Note that `specular_weight` is supposed to multiply the entire metallic lobe now, as in the snippet from the spec below. So in the graph, the most obvious implementation is to set the weight of the `generalized_schlick_bsdf` to `specular_weight` (if it has an overall weight, or otherwise just multiply the reflection tint by the ``specular_weight`).

image

reference/open_pbr_surface.mtlx
```diff
@@ -388,105 +404,101 @@
...
+      <input name="in2" type="color3" nodename="Ebase_Kcoat" />
+    </subtract>
+    <convert name="one_minus_Kcoat_color" type="color3">
+      <input name="in" type="float" nodename="one_minus_Kcoat" />
+    </convert>
+    <divide name="base_darkening" type="color3">
+      <input name="in1" type="color3" nodename="one_minus_Kcoat_color" />
+      <input name="in2" type="color3" nodename="one_minus_Ebase_Kcoat" />
+    </divide>
+    <multiply name="coat_weight_time_coat_darkening" type="float">
```
portsmouth, line 489: Should be `coat_weight_times_coat_darkening`

reference/open_pbr_surface.mtlx
```diff
@@ -580,10 +589,14 @@
     <luminance name="opacity_luminance" type="color3">
       <input name="in" type="color3" interfacename="geometry_opacity" />
```
portsmouth, line 590: Note that in the spec, `geometry_opacity` is defined to be a float. So it's unclear why it was set to be a color in the MaterialX inputs?

portsmouth:
Is a version of the MaterialX viewer available for download somewhere (i.e. a Windows binary) with 1.39 support?

reference/open_pbr_surface.mtlx
jstone-lucasfilm: Good catch, and that's very likely a leftover from the original definition of `opacity` in Standard Surface.  I can make the change here as part of our upgrade to MaterialX 1.39.

jstone-lucasfilm:
@portsmouth Since MaterialX 1.39 is still in development, we don't provide pre-built binaries for this unreleased version of MaterialX, so developers need to build it from the source in the `dev_1.39` branch:

https://github.com/AcademySoftwareFoundation/MaterialX/tree/dev_1.39

To maximize the visibility of OpenPBR, though, we've been updating the MaterialX Web Viewer to include each new release of OpenPBR, and it fully supports all of the functionality in MaterialX 1.39.

My hope is to publish a new release of OpenPBR as soon as we're happy with the latest behavior, and this will then become available in the web viewer as well.

reference/open_pbr_surface.mtlx
jstone-lucasfilm: That's a good note, and I can make that change as part of this upgrade as well.

jstone-lucasfilm:
Thanks for these notes, @portsmouth, and I believe I've addressed all of them in the latest implementation.

Here are a few renders of our example materials in the MaterialX Viewer, using the updated MaterialX 1.39 implementation in this pull request:

OpenPBR Aluminum Brushed:
OpenPBR_AluminumBrushed

OpenPBR Glass:
OpenPBR_Glass

OpenPBR Pearl:
OpenPBR_Pearl

portsmouth:
Making the `specular_weight` very large seems to produce a blown out render,  which shouldn't happen since the reflectivity should be clamped via this:
```xml
    <multiply name="scaled_specular_F0" type="float">
      <input name="in1" type="float" interfacename="specular_weight" />
      <input name="in2" type="float" nodename="specular_F0" />
    </multiply>
    <clamp name="scaled_specular_F0_clamped" type="float">
      <input name="in" type="float" nodename="scaled_specular_F0" />
      <input name="low" type="float" value="0.0" />
      <input name="high" type="float" value="0.99999" />
    </clamp>
```
So not sure why that is happening.

jstone-lucasfilm:
@portsmouth Since the default value for the `weight` input of `dielectric_bsdf` is already `1.0`, my thinking was that we should omit this duplicate statement in these node instances:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.PBRSpec.md#bsdf-nodes

Do you see a different result locally when you manually assign `1.0` as the weight for these nodes?  That would certainly be an unexpected result!

portsmouth:
`coat_darkening` works reasonably as expected, when turned on, e.g. here "matt wood" turns into "varnished wood".

No coat:

image

Coat (undarkened, default)

image

Coat (darkened, optional)

image


I think it would not be harmful to have the (more physically correct) darkening be the default.

jstone-lucasfilm:
@portsmouth I've fixed the legacy connection from specular_weight to the inputs of the dielectric_bsdf nodes, and the results look good in the MaterialX Viewer.

Let's take your additional suggestion for coat_darkening in a new conversation, as I think it's very reasonable, but I'd like to consider it separately from the current upgrade to MaterialX 1.39.

portsmouth:
In my view, the new coat roughening is also a good improvement to have applied by default (in this case, we already have this as the default, with no way to turn it off). It was subtly unconvincing when applying a rough coat to see the sharp specular highlight persist.

jstone-lucasfilm:
@portsmouth Agreed, we should decide upon a consistent approach to coat darkening and coat roughness in OpenPBR v1.0, whether that approach is "opt out" or "opt in", and that's a good discussion for the broader group.

portsmouth:
> Agreed, we should decide upon a consistent approach to coat darkening and coat roughening in OpenPBR v1.0, whether that approach is "enabled by default with opt out" or "disabled by default with opt in", and that sounds like a good discussion for the broader group.

Just to note, for coat roughening the spec just says it assumes the roughening is modeled somehow. It would be better to give a specific suggestion, which we do (plus the MaterialX implementation) in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/172. I suggest we get that into 1.0 as it more logically complete then, than leaving the implementation undefined.

There's no obvious way to turn the roughening effect off in a way which can be described physically. For example, in a Monte Carlo simulation of the layering, it would be impossible to avoid. So i'd leave that for later investigation (or perhaps never needed).

portsmouth (approved):
I think the Mltx reference could benefit from a bit more commentary in the file, e.g. stating how the calculations map to the equations in the spec, just for clarity. But we can certainly add that later as it would just be documentation.

LGTM!

PR #194, opened 2024-05-14
Slightly shift TOC closer to teaser image.

portsmouth:
Just a trivial tweak to reduce whitespace between the teaser images and the TOC:

image


jstone-lucasfilm (approved):
Ship it!

PR #195, opened 2024-05-14
Update contributors list.

portsmouth:
image


virtualzavie (approved):
I first thought it was inconsistent that Brecht van Lommel is sorted based on "van Lommel" while Eugene d'Eon is sorted based on "Eon".
But if I understand correctly [what Wikipedia says](https://en.wikipedia.org/wiki/Wikipedia:Categorization/Sorting_names), that's actually the proper way of sorting them.

jstone-lucasfilm (approved):
Ship it!

portsmouth:
We should also thank Anton Palmqvist for contributing his example materials.

virtualzavie:
Yes, if we could include them, that would look great.

PR #199, opened 2024-05-16
Add coat_darkening to default example

jstone-lucasfilm:
This changelist adds the coat_darkening input to our default material example, improving the visibility of this control in user testing.


virtualzavie (approved):
Good catch.

PR #203, opened 2024-05-28
Implement TIR fix in MaterialX

portsmouth:
Without fix:

image

With fix:

image

(In the `MaterialXView` app though, I don't see the difference. I'm not sure why, perhaps the original code is being pulled somehow?).


virtualzavie (approved):
Looks good to me.
Poking @peterkutz just to catch any error I would have missed.

jstone-lucasfilm (approved):

PR #204, opened 2024-05-28
Add uiname field to .mtlx nodedef

portsmouth:
Since the Maya team noted that they need this field in order to derive the node name (with correct capitalization and word break) to show in the editor:
>since the name of the node is open_pbr_surface, it will require custom code to turn it into "OpenPBR Surface".

>uiname metadata is already in the MaterialX specifications in the "[Custom Node Declaration NodeDef Elements](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#custom-node-declaration-nodedef-elements)" section for MaterialX 1.39 and I see this as well in the old 1.38 specification. Feel free to add it to OpenPBR surface and I will make sure LookdevX uses it.


jstone-lucasfilm (approved):
Thanks for this fix!

PR #205, opened 2024-05-28
Diffuse base_weight description changed to match MaterialX EON implementation

portsmouth:
As discussed in https://github.com/AcademySoftwareFoundation/MaterialX/pull/1822#issuecomment-2127738467, the MaterialX implementation assumed that the `base_weight` acts as an overall multiplier of the EON BRDF. While in the OpenPBR spec, we had `base_weight` instead multiplying `base_color`.

In fact the MaterialX intepretation seems better, since then the `base_weight` functions as a linear modulation of the overall albedo, rather than doing something non-linear and dependent on the color. This PR makes the necessary changes to the wording/math of the spec to implement this.

image

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -604,7 +604,9 @@ where $f_\mathrm{diffuse}$ is based on the classic Oren-Nayar model ([#Oren1994]
     ![Figure [rough_diffuse]: Zero roughness ($\sigma=0$, left) and maximum roughness ($\sigma=1$, right) diffuse materials with $\textbf{EON}$ BRDF.](dummy)
 </div>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1439,16 +1468,14 @@
 André Mazzone,
 Nikie Monteleone,
 Michael Nickelsky,
-Anton Palmqvist,
```
jstone-lucasfilm, line 1442: It looks like Anton and Nicolas were accidentally removed in this change!

portsmouth: Will fix

portsmouth: Done in 8821f630914bb6eb0f952970479c1f6d3067b690 (also added some others).

jstone-lucasfilm (approved):
This looks great, thanks @portsmouth!

Issue #149, opened 2023-12-17
question regarding color value range enforcement


Sam-Izdat:
Hi. Quite excited about the work you're doing.

Reading over the white paper, I noticed that the acceptable range for color values is set to be [0,1]. Does this imply that nonconforming color spaces should be excluded?
 
For example, ACES 2065-1, CIE LAB/LUV and OKLAB can have negative color values. Now, perceptual color spaces might be a questionable choice here, but it seems conceivable that someone would want to define their materials with AP0 primaries.

portsmouth:
This is a non-trivial question.

I would like to think of the color values in the model as defining the properties of the physical material. If the colors were specified in a fully spectral manner, I think they would correspond essentially to "spectral albedos", i.e. physically what fraction of each particular wavelength is absorbed or transmitted on passing through some layer, say. (Except for `emission_color` which would be the spectrum of emitted light).

Those albedos should physically all be in the range $[0,1]$ (i.e. it would be meaningless to absorb a negative amount of energy, or more than 100%).

However since we use RGB values, the spectral interpretation is ambiguous. The albedo per color channel represents some integral over (unknown) underlying physical albedos, with weights depending on the color space. I would expect that normally the channels should be in $[0,1]$, in standard color spaces and under some reasonable assumptions about the possible underlying spectral albedo curve. But I'm not sure, in the cases of the "non-conforming" color spaces you mentioned, perhaps it could make sense to allow that the RGB albedo components can be negative, or exceed 1? It would be strange to work with negative color channels in a renderer, but perhaps could be done consistently with care (obviously physical radiances must be positive, but RGB renderers don't work directly with physical radiance).

Generalizing this properly might also require re-expressing some of the math in the model (not to mention, in other similar models).  For example, in the standard Schlick model for metallic Fresnel:

![image](https://github.com/AcademySoftwareFoundation/OpenPBR/assets/2774364/f85521e2-bfa4-46d7-b6e7-29ef477afa00)

it is assumed that $\mathbf{F_0}=(1, 1, 1)$ corresponds to perfect Fresnel reflectivity. But in these exotic color spaces, is that the case?

KelSolaar:
Hello,

Thanks to @portsmouth for pointing me to this thread!

[...]

PR #208, opened 2024-05-28
Add footnote about colors being in-gamut (i.e. components in [0,1])

portsmouth:
Based on the interesting discussion in https://github.com/AcademySoftwareFoundation/OpenPBR/issues/149, it seems worth adding this side note:

image


portsmouth:
@anderslanglands and @KelSolaar Would appreciate your thoughts on whether this change makes sense.

KelSolaar:
It might also be:

"... all RGB colours are in-gamut of the working space, i.e., have components in the $[0, +\inf]$ range"

- "In-gamut" of what: We need to specify the container
- Why limiting to `1]`? Colour emission should not be restricted so we might need a case for albedo et al. and another one for emission.

portsmouth:
@KelSolaar 

>It might also be: "... all RGB colours are in-gamut of the working space, i.e., have components in the $[0, \infty]$ range"

OK, so in your view "in-gamut" (of the working color space of the shader) simply means that the color components are non-negative?

>"In-gamut" of what: We need to specify the container

Can you clarify what you mean by container? We say in the spec that "A color3 value is associated with a color space ACEScg by default)" and that "The assumed color space of all the color parameters .. by default .. is assumed to be ACEScg".  

So "in-gamut" is implied to mean in-gamut of that color space (or ACEScg if unspecified). Is that still under-specified in your view?

>Why limiting to 1? Colour emission should not be restricted so we might need a case for albedo et al. and another one for emission.

Yes right, so emission color should not be restricted to [0,1], but it seemed like according to our earlier [discussion ](https://github.com/AcademySoftwareFoundation/OpenPBR/issues/149) we had agreed that for all the other colors, which represent albedos or fractional tints, they should be restricted to [0,1] (regardless of the color space). Do you still agree with that?

KelSolaar:
> OK, so in your view "in-gamut" (of the working color space of the shader) simply means that the color components are non-negative?

Yes, e.g., HDR values are considered in-gamut of their colourspace for example.

> Is that still under-specified in your view?

I would certainly be very specific, there is an infinity of gamuts and it does not cost much to add.

> Do you still agree with that?

Yes!

portsmouth:
Sorry can you clarify what you mean by “there is an infinity of gamuts”?

We say the color space must be specified (or is ACEScg by default). How can we be more specific that that?

KelSolaar:
We are splitting hairs here but "gamut" by definition (Oxford Languages) is "the complete range or scope of something", what I'm saying is that the notes does not explicitly specifies it.

> Note that we assume that all RGB colors are in the specified color space's gamut, i.e. have components in the $[0, +inf]$ range.

This is better because the "something" is specified explicitly in the sentence right next to "gamut" and you do not have to reach the end of the sentence to understand what is being said.

In theory, the *i.e.,* portion of the sentence could be omitted and the sentence would be explicit by itself, I could only write: 

> Note that we assume that all RGB colors are in-gamut.

portsmouth:
@jstone-lucasfilm Note, don't merge this one yet. There needs to be some changes to the way we describe the color ranges in the spec.

portsmouth:
I modified this as follows:

image

image

image

portsmouth:
@jstone-lucasfilm can be merged?

portsmouth:
Note we added:
>The `emission_color` however is the exception which is permitted to have arbitrarily large (positive) components, since it represents an arbitrary luminance.

But the parametrization tables were not updated to make this change (despite the screenshots above?), so that is still to be done.

AdrienHerubel (changes requested):
Need a clean rebase on top of 1.1.1

portsmouth:
Can be closed, redone in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/260.

PR #211, opened 2024-05-30
Change thin film IOR default

portsmouth:
From 1.5 to 1.4.  

As this won't make much difference to the look in implementations that ignore the adjacent IORs of the film.

But for those that take it into account, this will make the film visible rather than invisible by default (since `specular_ior` is 1.5 by default, and `coat_ior` 1.6).


virtualzavie (approved):
Thank you for this fix.

jstone-lucasfilm:
Since MaterialX also depends on the thin-film IOR in rendering, we should make sure this change is reflected in the MaterialX reference implementation as well.

portsmouth:
> Since MaterialX also depends on the thin-film IOR in rendering, we should make sure this change is reflected in the MaterialX reference implementation as well.

Done in 60f67ef32d7c022daa51ea5e1f554d27c72eacd3

jstone-lucasfilm (approved):

PR #212, opened 2024-05-30
Move .html files from top level of repo to /html

portsmouth:
As it was too easy to mistakenly click on the html files instead of the "White paper" link to the spec.

(We had a complaint from someone who did that, and thought the resulting .md source was the formatted spec).

On merging this change, the GitHub Pages link **needs to be updated** to point to the new location of the .html (and verify that the .parametrization.md.html is also correctly linked to, from the front page and the spec). NB, only an admin can make the required changes to the GitHub Pages.


jstone-lucasfilm (approved):
Ship it!

PR #213, opened 2024-05-31
Update BibTeX citation (date, and capitalization)

portsmouth:
Updating the date to June.

Also forcing capitalization, so e.g. in journals (such as JCGT) which (annoyingly) render the titles by default as all lowercase, the capitalization of OpenPBR is retained, e.g. in JCGT:

image


virtualzavie (approved):
Good to know.

PR #215, opened 2024-06-04
Add a "resources" section to the front page

portsmouth:
With links to
  - MaterialX web viewer running OpenPBR default material
  - OpenPBR-viewer project and web app
  
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ OpenPBR Surface is an open standard hosted by the [Academy Software Foundation](
 
 ### Resources
 
-* [MaterialX Web Viewer](https://academysoftwarefoundation.github.io/MaterialX/?file=Materials/Examples/OpenPbr/open_pbr_default.mtlx) - WebGL   rasterization renderer using MaterialX [implementation](reference/open_pbr_surface.mtlx) of OpenPBR
+* [MaterialX Web Viewer](https://academysoftwarefoundation.github.io/MaterialX/?file=Materials/Examples/OpenPbr/open_pbr_default.mtlx) - WebGL rasterization renderer using MaterialX [implementation](reference/open_pbr_surface.mtlx) of OpenPBR
 * [OpenPBR-viewer](https://github.com/portsmouth/OpenPBR-viewer) - self-contained example implementation in a WebGL pathtracer (run [here](https://portsmouth.github.io/OpenPBR-viewer))
 * [#openpbr](https://academysoftwarefdn.slack.com/channels/openpbr) - public Slack channel for discussions, hosted by ASWF
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ OpenPBR Surface is an open standard hosted by the [Academy Software Foundation](
 
 ### Resources
 
-* [MaterialX web-viewer](https://academysoftwarefoundation.github.io/MaterialX/?file=Materials/Examples/OpenPbr/open_pbr_default.mtlx) - WebGL test renderer using MaterialX [implementation](reference/open_pbr_surface.mtlx) of OpenPBR
+* [MaterialX Web Viewer](https://academysoftwarefoundation.github.io/MaterialX/?file=Materials/Examples/OpenPbr/open_pbr_default.mtlx) - WebGL test renderer using MaterialX [implementation](reference/open_pbr_surface.mtlx) of OpenPBR
 * [OpenPBR-viewer](https://github.com/portsmouth/OpenPBR-viewer) - self-contained example implementation in a WebGL pathtracer (run [here](https://portsmouth.github.io/OpenPBR-viewer))
 * [#openpbr](https://academysoftwarefdn.slack.com/channels/openpbr) - public Slack channel for discussions, hosted by ASWF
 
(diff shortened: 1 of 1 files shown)


portsmouth:
@jstone-lucasfilm is this OK to merge?

README.md
```diff
@@ -15,9 +15,15 @@ OpenPBR Surface is an open standard hosted by the [Academy Software Foundation](
 
 * **[White paper](https://academysoftwarefoundation.github.io/OpenPBR/)**
 * **[Parameter reference](https://academysoftwarefoundation.github.io/OpenPBR/#parameterreference)**
-* **[Reference implementation](reference/open_pbr_surface.mtlx)** – written in [MaterialX](https://materialx.org/)
+* **[MaterialX reference implementation](reference/open_pbr_surface.mtlx)**
 * **[BibTeX citation](openpbr.bib)**
 
+### Resources
+
+* [MaterialX web-viewer](https://academysoftwarefoundation.github.io/MaterialX/?file=Materials/Examples/OpenPbr/open_pbr_default.mtlx) - WebGL test renderer using MaterialX [implementation](reference/open_pbr_surface.mtlx) of OpenPBR
```
jstone-lucasfilm, line 23: Just a minor note, but for consistency with other documentation, I'd recommend referring to this application as "MaterialX Web Viewer".

portsmouth: Done in 41271ff0a30539559b90b7de83f1640634c9952f.

README.md
```diff
@@ -15,9 +15,15 @@ OpenPBR Surface is an open standard hosted by the [Academy Software Foundation](
 
 * **[White paper](https://academysoftwarefoundation.github.io/OpenPBR/)**
 * **[Parameter reference](https://academysoftwarefoundation.github.io/OpenPBR/#parameterreference)**
-* **[Reference implementation](reference/open_pbr_surface.mtlx)** – written in [MaterialX](https://materialx.org/)
```
jstone-lucasfilm, line 18: In some ways, I still prefer the original wording, which makes it clear that there's just a single reference implementation of OpenPBR.  Maybe we can consider this proposed change further in the future, and focus on the Resources section for this change?

README.md
```diff
@@ -15,9 +15,15 @@ OpenPBR Surface is an open standard hosted by the [Academy Software Foundation](
...
+* **[MaterialX reference implementation](reference/open_pbr_surface.mtlx)**
 * **[BibTeX citation](openpbr.bib)**
 
+### Resources
+
+* [MaterialX Web Viewer](https://academysoftwarefoundation.github.io/MaterialX/?file=Materials/Examples/OpenPbr/open_pbr_default.mtlx) - WebGL test renderer using MaterialX [implementation](reference/open_pbr_surface.mtlx) of OpenPBR
```
jstone-lucasfilm, line 23: I'd recommend phrasing this as "WebGL rasterization renderer", making a clearer contrast with the "WebGL pathtracer" below.

README.md
portsmouth: I don't think there really is (or should be) an official "reference implementation", as the spec itself is supposed to define how to implement the model. The MaterialX implementation is also not really currently a good reference in my opinion, as it's both hard to understand and also uses a bunch of approximations (for example, my implementation is the viewer is actually more faithful to the spec).

Even so, it's the only official reference implementation we have, so we should flag it as such in the main "Specification" section. I don't see a problem with referring to it as "MaterialX reference implemention", not *the* reference implementation.

jstone-lucasfilm: @portsmouth This sounds like a deeper discussion that you and I should have, and I'd recommend that we separate this from the current set of proposed updates to our front page.  :)

portsmouth: OK, so the change you want is just that we put back
>* **[Reference implementation](reference/open_pbr_surface.mtlx)** – written in [MaterialX](https://materialx.org/)

Correct?  If so, I think the difference doesn't matter much, so fine. (It's "a" reference implementation, not "the" one, from my point of view, but we don't say either so no problem).

jstone-lucasfilm: Yes, I'd just recommend leaving this text alone for now, since it's served us well up to this point.  If you're interested in reframing our approach to reference implementations in OpenPBR, then we should discuss that directly (e.g. in TSC meetings or on Slack), and we can make adjustments to our framing based on the findings of that discussion.

portsmouth: I don't think i'm proposing a reframe.. I was just stating it the way it is - we don't say in the spec that the MaterialX implementation is the ground truth. Maybe there's some difference of opinion about what a "reference implementation" means then. I'm OK with the leaving the text as is though.

portsmouth:
Fixed up in 97570e402cd3566495a7bca489a02463c3532792,  to look as below:

image

README.md
```diff
@@ -18,6 +18,12 @@ OpenPBR Surface is an open standard hosted by the [Academy Software Foundation](
 * **[Reference implementation](reference/open_pbr_surface.mtlx)** – written in [MaterialX](https://materialx.org/)
 * **[BibTeX citation](openpbr.bib)**
 
+### Resources
+
+* [MaterialX Web Viewer](https://academysoftwarefoundation.github.io/MaterialX/?file=Materials/Examples/OpenPbr/open_pbr_default.mtlx) - WebGL   rasterization renderer using MaterialX [implementation](reference/open_pbr_surface.mtlx) of OpenPBR
```
jstone-lucasfilm, line 23: This is looking close to the mark, @portsmouth, though I still see some weird spacing in the text on this line.

portsmouth: Fixed in 1ca6c3b27165a5b136758c59a808e8e106371f16

jstone-lucasfilm (approved):
Ship it!

virtualzavie (commented):
Nice, thanks for the useful addition.

PR #216, opened 2024-06-05
Update OpenPBR default example

jstone-lucasfilm:
This changelist updates the OpenPBR default example, matching its values to the latest default values of the shading model.

### DIFF
--- a/examples/open_pbr_default.mtlx
+++ b/examples/open_pbr_default.mtlx
@@ -20,7 +20,7 @@
     <input name="transmission_scatter" type="color3" value="0, 0, 0" />
     <input name="transmission_scatter_anisotropy" type="float" value="0.0" />
     <input name="transmission_dispersion_scale" type="float" value="0.0" />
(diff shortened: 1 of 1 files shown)


portsmouth (approved):

examples/open_pbr_default.mtlx
```diff
@@ -35,8 +35,8 @@
     <input name="coat_roughness_anisotropy" type="float" value="0.0" />
     <input name="coat_ior" type="float" value="1.6" />
     <input name="coat_darkening" type="float" value="1.0" />
-    <input name="thin_film_weight" type="float" value="0.0" />
-    <input name="thin_film_thickness" type="float" value="0" />
+    <input name="thin_film_weight" type="float" value="0" />
```
virtualzavie, line 38: Not that it matters, but it looks like the example is inconsistent when it comes to including or not the `.0`.
In the specification and in the parameter list in particular, I was careful to remain concise for readability, but examples are a different context.

PR #217, opened 2024-06-11
Enable Zeltner sheen

jstone-lucasfilm:
This changelist enables Zeltner sheen in the reference implementation of OpenPBR, leveraging the new functionality in MaterialX 1.39.

Additionally, the open_pbr_velvet.mtlx example has been updated to account for the visual differences between Conty-Kulla and Zeltner sheen.


theblackunknown (approved):
LGTM
Which MaterialX version/commit is reaquired to be in sync with this change ?
Or maybe it does not matter because those changes are already replicated inside AcademySoftwareFoundation/MaterialX ?

examples/open_pbr_velvet.mtlx
```diff
@@ -8,6 +8,6 @@
     <input name="specular_roughness" type="float" value="0.8" />
     <input name="fuzz_weight" type="float" value="1" />
     <input name="fuzz_color" type="color3" value="0.4, 0.4, 0.4" />
-    <input name="fuzz_roughness" type="float" value="0.2" />
+    <input name="fuzz_roughness" type="float" value="0.5" />
```
theblackunknown, line 11: Is this empirical, or is there a closed form to fit Conty-Kulla on Zeltner one ?

jstone-lucasfilm: I'm not aware of a closed-form solution, but we can definitely put together a good approximation and include it in a future shader translation graph from `standard_surface` to `open_pbr_surface`.

For now, I'm just estimating the visual equivalent of the original settings with Zeltner sheen, in order to roughly maintain the look of the example material in the new model.

jstone-lucasfilm:
> Which MaterialX version/commit is reaquired to be in sync with this change ? Or maybe it does not matter because those changes are already replicated inside AcademySoftwareFoundation/MaterialX ?

This will match the very first release of MaterialX 1.39, which is technically still in development, with a release planned for the end of this month.

Issue #144, opened 2023-11-14
subsurface_radius_scale should maybe be a color (not a vector3)?


portsmouth:
Currently we say that `subsurface_radius_scale` has type `vector3`.  But what does this mean operationally?  It's confusing for people doing UI, as to whether this parameter should have a color picker, but we certainly want it to. The components are in [0,1] range so there's no issue with the values.

Perhaps it just means that the parameter should not be color managed, but do we not want that, as the meaning of the RGB channels (i.e. which "curve" of wavelength bands a given radius should apply to)  should also change as the color space changes?  So i'm not sure it doesn't make more sense to just leave it as a color.

![image](https://github.com/AcademySoftwareFoundation/OpenPBR/assets/2774364/bb77a82e-582a-476e-83bb-6d2c06345cf6)

jstone-lucasfilm:
Thanks for posting this issue, @portsmouth, and this has been resolved in #220.

PR #220, opened 2024-06-25
Update subsurface color types

jstone-lucasfilm:
This changelist updates the types associated with physical color values for subsurface scattering in OpenPBR, aligning with the conclusions of recent threads on ASWF Slack channels.

- Change `subsurface_radius_scale` from a `vector3` to a `color3` in the specification, aligning with the MaterialX implementation of OpenPBR.
- Change the `radius` input of `subsurface_bsdf` from a `vector3` to a `color3` in the MaterialX implementation, aligning with the current definition of the `subsurface_bsdf` node in MaterialX 1.39.

### DIFF
--- a/index.html
+++ b/index.html
@@ -750,7 +750,7 @@ Subsurface params                   | Label        | Type      | Range          
 **`subsurface_weight`**             | Weight       | `float`   | $ [0, 1]      $ |            | $ 0                $ | Mix weight between subsurface and diffuse slabs
 **`subsurface_color`**              | Color        | `color3`  | $ [0, 1]      $ |            | $ (0.8, 0.8, 0.8)  $ | The observed reflection color of $V^\infty_\mathrm{subsurface}$
 **`subsurface_radius`**             | Radius       | `float`   | $ [0, \infty) $ | $ [0, 1] $ | $ 1                $ | Length scale of MFP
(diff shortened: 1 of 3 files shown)


portsmouth:
This accords with what I proposed in https://github.com/AcademySoftwareFoundation/OpenPBR/issues/144. 

As we discussed on Slack, there is still some work to do to understand how/whether color management of the `subsurface_radius_scale` can be made to work better, as due to the non-linearity of the relationship between the radius (which is the RGB mean-free-path) and the observed color, the render appearance can change significantly on changing color space.

But it still makes more sense for this to be a color than a `vector3`, as it needs some kind of color management (and is picked as a color, driven by RGB texture, etc.).

virtualzavie (approved):
I haven't followed the discussion on this topic, but the proposed change looks good to me.

portsmouth:
Just to note, we also don't currently make it very clear in the OpenPBR spec what it means that a quantity is a `color` versus a `vector3`. We should probably say explicitly that we assume that the `color3` quantities are color managed in the standard way.

jstone-lucasfilm:
@portsmouth In this situation, we can just fall back on the specified behavior of color spaces in MaterialX, which is inherited by OpenUSD and NanoColor as well:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#color-spaces-and-color-management-systems

portsmouth:
@jstone-lucasfilm sure, though it deserves at least a mention in the OpenPBR spec, that we expect some such logic to be applied by the host application.

PR #236, opened 2024-09-03
Image overhaul

portsmouth:
Updating images to use the new shaderball, and adding some new images.

Also indenting the anisotropy diagram to make it less distracting.

image


index.html
```diff
@@ -1000,6 +1004,11 @@
 
 The base albedo $E_b$ which appears in equation [general_darkening_formula] represents the normal-incidence albedo of the entire base beneath the coat. This albedo can be approximated as the normal-incidence albedos of the individual slabs of the base, blended according to their mix weights.
 
+![](images/coat_darkening_0.0.png width=99%) ![](images/coat_darkening_0.5.png width=99%) ![](images/coat_darkening_1.png width=99%)
+<div class="shifted-caption">
+    ![Figure [fuzz]: A checkerboard **`coat_weight`**, varying the **`coat_darkening`** over 0, 0.5, 1 (default).](dummy)
```
virtualzavie: I find this particular illustration difficult to understand. The wood looks glossy everywhere, which I interpret as coated everywhere. If the intent is to show a checker of coated vs non coated parts, I'd suggest using a rough base layer.

AdrienHerubel (changes requested):
Let's address the comment from @virtualzavie and get that change in for the fix release, we can add more updates in other PRs.

index.html
```diff
@@ -1038,6 +1042,11 @@
...
+</div>
+
```
AdrienHerubel:
```suggestion
```

AdrienHerubel (approved):
Let's get this first batch of improved images in. I removed the image mentioned by @virtualzavie as being hard to understand.

PR #240, opened 2024-10-22
MaterialX: metal edge tint should not be multiplied by specular_weight

portsmouth:
Fixes a bug in the MtlX graph. The edge-tint color for the `generalized_schlick_bsdf` should not be multiplied by `specular_weight`, as the whole lobe is multiplied by `specular_weight` as well.

(Note that MtlX still does the wrong thing if  `specular_weight` > 1, but that's not fixable on the OpenPBR side, unless we just clamp the weight so `specular_weight` > 1 does nothing to the metal).


jstone-lucasfilm (approved):

Just to double-check, do we want to include this in `dev_1.2` or in `dev_1.1.1`?

portsmouth:
It seems harmless to include it in `dev_1.1.1`, assuming that means a bug-fix release. It would be look breaking, but presumably in a good way.

jstone-lucasfilm:
For now, let's merge this to its targeted 1.2 branch, and we can integrate it back to 1.1.1 if needed.

Issue #241, opened 2024-10-29
Realtime use case affordances


dgovil:
We'd discussed this in a few other meetings, but I'd like to formally request some affordances in OpenPBR for realtime use cases, where realtime could include:

1. Consoles
2. Mobile devices
3. Web

The primary things that come to mind:

1. It's difficult for many rasterizers to support some of the features like transmission and subsurface. It would be great to allow for the following:
    - Language that says how free implementations are with regards to supporting different feature sets. Right now its not completely clear whether an implementation can approximate or elide features that aren't possible on a specific hardware target
    - Suggestions for how to approximate features if full path traced support is not possible

2. Realtime renderers often desire ambient occlusion maps which is not something that OpenPBR supports. It would be great to have some kind of officially supported way to do occlusion for renderers that prefer that route.

portsmouth:
We do say:

>... we want this material model to be renderable on a wide range of platforms, from offline path tracers all the way to real-time game engines on mobile devices. Enforcing a particular implementation would make the use of the material model impractical for certain classes of renderers, and ultimately make the model less useful. For this reason we consider the choice of a specific implementation of the final BSDF to be outside the scope of this specification.

So we could try to be more explicit, but it seems fairly clear that implementations can really do whatever they want (or have to given the constraints) -- the only penalty is that they will not match other renderers very well, if they deviate from the spec. Which maybe is fine for that particular use case, e.g. no-one is going to expect a mobile game engine to match Arnold.

Adding suggestions for how a game engine can approximate refraction or subsurface scattering seems a bit beyond the scope of the spec. Though perhaps we could at least mention that this is a practical problem, and suggest some references or resources. Also, there are example implementations available that do some such approximations for real-time (e.g. MaterialX, and Three.js is in progress), though I think we could provide more clear links to those.

I also think we should provide suggested translations to some other popular shading models (e.g. Standard Surface, Disney Principled. USDPreviewSurface, GLTF) in some easily portable form, which would be a practical solution for rasterizers if they already support one of those models.

meshula:
I'd argue that we need to provide strong guidance for implementors, as opposed to diverting them to other BSDFs.

If someone asked me, hey Nick get this running on a phone on 30 characters, the first thing I'd do is draw a graph of OpenPBRs components and how they compose and both mark that graph, and create a matrix of cost versus subjective impact, and I'd populate that matrix indications

- omission breaks the universe
- omission detectable by experts
[...]

PR #248, opened 2025-02-04
Add "Flexibility of implementation" section

virtualzavie:
This PR is an attempt at addressing requests for some guidance from the real-time community, including #241.

The proposal is to state clearly, in one place that is easy to identify and refer to in the document, what is expected from implementations and what they are allowed to do while still complying to the specification.

Ideally, an annex would propose some approximations they can use, with code snippets ready to use.

image


approximation.md.html
```diff
@@ -0,0 +1,55 @@
+<meta charset="utf-8" emacsmode="-*- markdown -*-">
+<link rel="stylesheet" href="style/style.css">
+<meta name="viewport" content="width=device-width,initial-scale=1">
+
+
+Approximations for constrained implementations
+==============================================
+
+This specification describes a model which includes a number of features that can be computationnally expensive and therefore impractical to support in certain contexts, like real-time applications or mobile devices.
```
dgovil: Typo: Computationnally -> Computationally (double n)

approximation.md.html
```diff
@@ -0,0 +1,55 @@
...
+
+The Historical background and objectives section explains how the goal of this specification is to provide an ideal, reference appearance. The implementer is, however, free to make decisions and approximations to satisfy their practical constraints. We cannot dictate what those approximations should be, as this depends on the use case. We can, however, propose some possible approximations and offer guidelines to help the implementer make those decisions.
+
+As a general rule of thumb, we invite the implementer to think of those approximations in the same way one would think of different levels of detail (LOD) variants of an asset. Just like such an asset would become more coarse with distance while retaining as much detail as necessary to remain faithful to the original, an implementation of OpenPBR may decide to make simplifications, as long as the result remains reasonably faithful to the intent given the practical constraints. In the most extreme hypothetical case, the model might even get reduced to as little as a single Lambert BRDF.
```
dgovil: Typo: more coarse -> coarser

approximation.md.html
```diff
@@ -0,0 +1,55 @@
...
+
+This means the implementer must carefully consider what to do with parameters they do not plan to fully support. A parameter that drastically modifies the appearance of a material should not be discarded just because the corresponding effect is not supported. As an example, in a renderer that doesn't have subsurface scattering, the subsurface color might be used as the albedo of a diffuse BRDF.
+
+In the following sections we propose possible approximations for the different components of the OpenPBR Surface material.
+
+
+Layering and mixing
+-------------------------------------
+
+As mentioned in the Reduction to a mixture of lobes section, the mix and layer operations can approximated as a weighted sum of BSDF lobes.
```
dgovil: Typo: approximated -> approximate or be approximated

dgovil (approved):
Thanks for explicitly calling out that approximations are permitted. This will be a very useful clarification.

I left a couple small typo notes as I was going through.

approximation.md.html
virtualzavie: Thank you, fixed at [8416388](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/248/commits/8416388ee9abe5cd1ae7be1ef5cc9c9dfab754f4).

approximation.md.html
virtualzavie: Fixed at [8416388](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/248/commits/8416388ee9abe5cd1ae7be1ef5cc9c9dfab754f4).

approximation.md.html
virtualzavie: Fixed at [8416388](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/248/commits/8416388ee9abe5cd1ae7be1ef5cc9c9dfab754f4).

approximation.md.html
```diff
@@ -0,0 +1,55 @@
...
+<link rel="stylesheet" href="style/style.css">
+<meta name="viewport" content="width=device-width,initial-scale=1">
+
+
+Approximations for constrained implementations
+==============================================
+
+This specification describes a model which includes a number of features that can be computationally expensive and therefore impractical to support in certain contexts, like real-time applications or mobile devices.
+
+The Historical background and objectives section explains how the goal of this specification is to provide an ideal, reference appearance. The implementer is, however, free to make decisions and approximations to satisfy their practical constraints. We cannot dictate what those approximations should be, as this depends on the use case. We can, however, propose some possible ones and offer guidelines to help the implementer make those decisions.
```
meshula: suggest: We don't dictate

obviously we could dictate, but choose not too

meshula (commented):
this PR seems to include much unrelated history. does it need to be rebased?

approximation.md.html
virtualzavie: Thank you. Reworded at [fdd20b3](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/248/commits/fdd20b3bb680eb9e5afc12a9efed045a6bd170c0).

virtualzavie:
> this PR seems to include much unrelated history. does it need to be rebased?

Yes, most likely. I think I either didn't branch off the right point, or didn't open the PR towards the proper branch.

virtualzavie:
> this PR seems to include much unrelated history. does it need to be rebased?

I've rebased the branch on top of main, excluding the commits from the 1.x branches that were unrelated to the PR.

meshula:
Thanks! The diff is so much easier to look at now :)

approximation.md.html
```diff
@@ -0,0 +1,55 @@
...
+
+This means the implementer must carefully consider what to do with parameters they do not plan to fully support. A parameter that drastically modifies the appearance of a material should not be discarded just because the corresponding effect is not supported. As an example, in a renderer that doesn't have subsurface scattering, the subsurface color might be used as the albedo of a diffuse BRDF.
+
+In the following sections we propose possible approximations for the different components of the OpenPBR Surface material.
+
+
+Layering and mixing
+-------------------------------------
+
+As mentioned in the Reduction to a mixture of lobes section, the mix and layer operations can be approximated as a weighted sum of BSDF lobes.
```
meshula: since `Reduction to a mixture of lobes` is not a named section in this document, can we link or otherwise direct the reader somehow?

approximation.md.html
```diff
@@ -0,0 +1,55 @@
...
+
+### Anisotropic case
+
+Several solutions are possible for the anisotropic case:
+
+- [#Cocco2024] recently proposed an image-based lighting approximation of anisotropic specular reflection consisting in several samples along the main BRDF axis.
+
+- [#McAuley15] describes a simpler and cheaper method based on [#Revie2011], consisting in modifying normals and offering a very crude approximation of anisotropic reflection.
+
+- Finally, the crudest approximation would be to ignore entirely the `specular_roughness_anisotropy` for the base layer specular lobe (respectively `coat_roughness_anisotropy` for the coat specular lobe), as the anisotropy parametrization in designed to preserve the average roughness regardless of the anisotropy (see the Microfacet model section for details).
```
meshula: since the `Microfacet model section` is not a named section in this document, can we link or otherwise direct the reader somehow?

meshula (commented):
This is looking good to me. I've suggested that we should more explicitly direct the reader to the named sections ~ but maybe that's not necessary because the document gets published in a way where the sections are immediately obvious?

portsmouth:
As proposed in the meeting, perhaps we can repurpose this into two PRs, with:
 - just the wording change that permits approximations, more explicitly. This could maybe just go into the introduction?
 - a separate document that will be the skeleton of the guidelines?  This could also move chunks that are essentially just approximations, into the guidelines. Some candidates for moving out of the spec are (these might be better as sub-PRs though):
   * the "Reduction to a mixture of lobes" section, is essentially detailing derivation of one approximate form of the model, that looks like Standard Surface. (This is the main task arguably for an implementer -- i.e. mapping the layer structure to a weighted BSDF combination).  It would be interesting to know for example the process that Adobe took, to map the OpenPBR layer structure to a concrete implementation with individual BSDFs.
   * the coat darkening/roughening/TIR math,  that amounts to a choice of approximation. (Not the part that specifies what the darkening parameter does).
   * more SSS remapping choices could be stated.

virtualzavie:
I've reduced this PR to the wording change, with a section in `index.html`.
I've removed the guidelines part, and will reintroduce it with a separate PR.

approximation.md.html
meshula: (this is resolved)

meshula:
What's the current status? The latest revisions look good from my perspective. I'm looking forward to the proposal being merged.

index.html
```diff
@@ -1431,6 +1427,20 @@
 We provide a [reference implementation](reference/open_pbr_surface.mtlx) in MaterialX, which is based on the derivation in the previous section and the implementation of Autodesk Standard Surface [#Georgiev2019].
 
 
+Flexibility of implementation
+=============================
+
+We want this material model to be renderable on a wide range of platforms, from offline path tracers all the way to real-time game engines on mobile devices.
```
portsmouth, line 1433: Not sure this really needs to be a separate section. It could just be paragraphs added to the "objectives" section, so the high-evel objectives/policies are all stated in one place.

index.html
```diff
@@ -1431,6 +1427,20 @@
...
+
+The Historical background and objectives section explains how the goal of this specification is to provide an ideal, reference appearance. Complete conformance to the specification is defined as reproducing all the physical light transport effects of that ideal appearance.
+
+However, a number of features can be computationally expensive and therefore impractical to support under certain constraints. Enforcing a specific implementation would tailor the material model to a particular set of constraints, limiting its versatility, which would ultimately make it less useful. For this reason, the specification focuses solely on defining the target appearance. The choice of the final BSDF implementation and its associated trade-off is left entirely to the implementer, allowing them to best suit it to their needs.
```
portsmouth, line 1437: Maybe worth detailing what "constraints" means in this context. Presumably hardware constraints, like CPU/GPU speed and memory? What else, things like choice of graphics API/language?

index.html
```diff
@@ -1431,6 +1427,20 @@
...
+
+Each implementation must decide what level of approximation to use, trading off accuracy for efficiency according to its own particular use case. As a guideline, we suggest approaching those approximations conceptually like different level of detail (LOD) variants of an asset. Just like such an asset would become coarser with distance while retaining as much detail as necessary to remain faithful to the original, an implementation of OpenPBR may decide to make simplifications, as long as the result remains reasonably faithful to the intent given the practical constraints. In the most extreme hypothetical case, the model might even get reduced to as little as a single Lambert BRDF.
+
+This means the implementer must carefully consider the handling of parameters they do not plan to fully support. A parameter that drastically modifies the appearance of a material should not be discarded just because the corresponding effect is not supported. As an example, in a renderer that doesn't have subsurface scattering, the subsurface color might be used as the albedo of a diffuse BRDF.
```
portsmouth, line 1441: The general quantitative approach would be to try to make an "as faithful as possible" mapping from the full OpenPBR model to whatever the "target" model is. 

For example if the target is the Lambert model, there is one parameter to fit -- the albedo color. The natural thing to do is then set that color to the "diffuse albedo" color of the full OpenPBR model, which would be sum of the albedos of the various lobes which are considered to be diffuse (i.e. rough) enough, e.g. diffuse/subsurface/fuzz, and metal (if rough). That would then automatically account for e.g. the subsurface color, with the correct subsurface weight.

A function which computes those albedos could look something like [this](https://github.com/portsmouth/OpenPBR-viewer/blob/39c32893f19fe31cb02ae073053c9e68865f4bc7/glsl/rasterization/openpbr.frag.glsl#L803).

>As an example, in a renderer that doesn't have subsurface scattering, the subsurface color might be used as the albedo of a diffuse BRDF.

That is less clear, since in general you want to use a mix of the colors (i.e. albedos) of the various lobes, according to the lobe weights.

index.html
virtualzavie: I would argue this consideration is important enough to warrant its own section[1] and appear in the table of contents, as it directly addresses an entire industry.

One reason that prompted this discussion is that even though the information was present to an extent, it was not obvious without reading through the document. As a section, it is easy to locate for the reader, and also indicates to future contributors where remarks on this topic should be, to prevent them from being scattered.

1: section or sub-section, if we prefer a different ordering or structure.

index.html
virtualzavie: I like the idea of reasoning with albedos. I'll do another pass.

AdrienHerubel (approved):
Ready to be merged.

PR #251, opened 2025-03-07
MaterialX: fix bug in thin-walled subsurface

portsmouth:
As pointed out by Brian Sharpe of SideFX, there is a bug in the thin-walled subsurface, since `specular_color` is accounted for in both the BSDF weights and as an explicit multiplier.

Also, the graph can be simplified a bit using the mix operation with a more carefully chosen weight.


jstone-lucasfilm:
Sorry for the spam above, @portsmouth, which was my attempt to rebase the change to `dev_1.1.1`.  For now I'll leave this as a change in `dev_1.2`, and feel free to update it if you'd like to target a different branch.

jstone-lucasfilm (approved):

Just to double-check, do we want to include this in `dev_1.2` or in `dev_1.1.1`?

jstone-lucasfilm:
For now, let's merge this to its targeted 1.2 branch, and we can integrate it back to 1.1.1 if needed.

PR #258, opened 2025-05-30
Subsurface in thin-walled mode, small clarification

portsmouth:
I noticed that the `subsurface_radius` and `subsurface_radius_scale` are just ignored in the thin-walled mode of the subsurface, which should be said explicitly otherwise potentially it's confusing.

image


AdrienHerubel (changes requested):
Need a clean rebase to 1.1.1

peterkutz (approved):
This is a helpful clarification.

On a related note, I recently noticed a few other aspects of thin-walled mode that could also potentially benefit from clarification:

https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1749690787065219

Issue #246, opened 2025-01-30
Add more implementation guidance for thin-film iridescence


peterkutz:
Think-film iridescence can be difficult to implement, and the paper we reference can make it even more complex due to advanced color-related transformations. This task is to add more implementation guidance to the OpenPBR spec based on our own implementations of thin-film iridescence, which are summarized on Slack:

Thin-film GLSL implementation (from @portsmouth ):
https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1710080253470669

Thin-film iridescence in Eclair (from @peterkutz ):
https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1732332598108709?thread_ts=1710080253.470669&cid=C06365EAQMP

More thin-film implementation notes: Improved iridescence on metals (from @peterkutz ):
https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1733848688599819?thread_ts=1710080253.470669&cid=C06365EAQMP

PR #262, opened 2025-06-26
Revise and expand thin-film iridescence documentation

peterkutz:
This PR revises and expands the thin-film iridescence section of the OpenPBR documentation. It includes:

- A new paragraph describing the _locally spectral_ approach to modeling thin-film interference, as described on Slack (especially [here](https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1710080253470669), [here](https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1732332598108709?thread_ts=1710080253.470669&cid=C06365EAQMP), and the series of messages starting [here](https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1733848688599819?thread_ts=1710080253.470669&cid=C06365EAQMP)).
- Additional description of the originally recommended Belcour–Barla approach so that the locally spectral approach can easily be compared and contrasted with it.
- Clarifications to the explanation of iridescence itself and the effects of the parameters.
- Updated organization of the section to better distinguish between modeling approaches and shared considerations.
- Improvements to wording for readability, conciseness, and accuracy.

This PR addresses OpenPBR issue https://github.com/AcademySoftwareFoundation/OpenPBR/issues/246

Below are images comparing the current and proposed versions of this section.

**CURRENT VERSION**

thin-film_section_CURRENT

**PROPOSED VERSION**

thin-film_section_PROPOSED

As you can there are modifications across the entire section, as mentioned in the overview at the top of this PR. The most substantive new addition to the section is the new paragraph highlighted below.

**CLOSE-UP OF BRAND NEW PARAGRAPH IN PROPOSED VERSION**

brand_new_paragraph


portsmouth:
All sounds good. I assume we'll be expanding on this "locally spectral approach" a bit more in the tech report.

Another related thing I looked into recently was the energy conservation of the thin-film. We had a report from the Redshift team that "metalness with thin-film looks 'neon' in Arnold". I assumed that this might be a bug, since the thin-film should never generate energy. But had a debate internally about whether actually due to wave interference perhaps effectively the albedo can exceed 1 (maybe via some sharing of energy between frequency bands) leading to a glow effect, so maybe we were seeing the correct result.

This seemed impossible to me physically, since each wavelength is independent, and the interference cannot produce more energy out at that wavelength than was input.

I wrote a script which explicitly computes the albedo of the film as a function of input direction and wavelength, and indeed it is always <= 1. In the special case of a perfectly reflecting metallic base (with Fresnel 1), the energy is perfectly conserved and the effect of the film is only a phase shift. (This corresponds to what happens in a [Gires–Tournois etalon](https://en.wikipedia.org/wiki/Gires%E2%80%93Tournois_etalon) as pointed out by a colleague). Thus indeed a metal with white F0 and tint, plus an arbitrary thin-film, should pass a furnace test (as we say in the spec).

(The "neon" look in Arnold was, it turns out, due to a small bug in our implementation of the Barla paper).

It might be interesting to show some plots of this in the tech report.

peterkutz:
Thanks for reviewing @portsmouth .

Yes, I think we can expand more on this "locally spectral approach" in the tech report, but it seems out of scope here. I think we can also provide the key physics formulas in the tech report.

Yes, that is an important point. Interference should only change the ratio of light reflected and transmitted. The thin film itself should neither add nor remove energy. Any apparent energy gain/loss or color shift comes from this energy redistribution and different amounts of absorption by the base surface. I think that's actually important and nonobvious enough to warrant a note in the spec, so I added a note in [a new commit](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/262/commits/d046bf9fec8ef6a88dd9b3f91fc5480bb98a4234):

> Because the thin film is non-absorbing and interference-based, it only redistributes the probabilities of reflection and transmission; therefore, it should not violate energy conservation.

AdrienHerubel (changes requested):
This PR was discussed in the September 16 meeting, there was a consensus that we should not recommend techniques that do not have a readily available implementation in the spec. We also said that the Belcour technique is the de-factor standard and we should keep the wording stating so.

peterkutz:
Thanks @AdrienHerubel. Agreed, and I’ve updated the PR accordingly.

The new wording keeps Belcour & Barla as the official recommendation / de-facto standard (especially for RGB rendering contexts), and no longer frames any alternative as something the spec "recommends".

What I did keep is a brief allowance that spectral renderers may implement thin-film directly from the underlying Fresnel + Airy interference equations (i.e., a first-principles evaluation per wavelength with integration in the renderer’s spectral pipeline). This isn’t presented as a new technique or a separate "model"; it’s just the direct physics that B&B starts from as well. The updated text does not describe implementation details, so it shouldn’t steer anyone toward an unverified formulation. It also explicitly states that such implementations should be validated against the B&B reference in the regimes where B&B applies. Allowing a first-principles spectral implementation (without endorsing it) is also consistent with the spec's description of dispersion, and with the spec's overall philosophy of describing a physical configuration that can be implemented in the most accurate way possible.

Here's the final state of the thin-film section:
NewThinFilmSection

And here's a closeup of the part that positions B&B as the official approach but allows a first-principles spectral approach when applicable:
NewThinFilmSectionCloseup

portsmouth:
I think the proposed minimal wording is fine. Ideally, @peterkutz  I think we should write-up in more detail how the locally spectral approach works, as a separate project which we can then reference (which is interesting in its own right, as a potentially superior method to the B&B paper). But this is fine for the interim.

AdrienHerubel (approved):
This PR was extensively discussed and there is a consensus that the changes make it easier to implement thin-film. This PR is recommended to be backported to 1.1.1

PR #265, opened 2025-07-08
Integrate OpenPBR update from MaterialX project

jstone-lucasfilm:
This changelist integrates one post-1.1 update to OpenPBR Surface from the MaterialX project:

- Add code generation hints support (https://github.com/AcademySoftwareFoundation/MaterialX/pull/1954)

### DIFF
--- a/reference/open_pbr_surface.mtlx
+++ b/reference/open_pbr_surface.mtlx
@@ -391,11 +391,7 @@
       <input name="bg" type="BSDF" nodename="opaque_base" />
       <input name="mix" type="float" interfacename="transmission_weight" />
     </mix>
-    <invert name="thin_film_weight_inv" type="float">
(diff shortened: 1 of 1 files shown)


AdrienHerubel (changes requested):
The optimized version of the graph was reverted in the default version of the graph, and the specialized version of the graph is only used in specific targets. I suggest we keep in sync with the default version of the graph.

jstone-lucasfilm:
@AdrienHerubel This PR is now in sync with the default version of the OpenPBR graph in the MaterialX project, and doesn't contain any platform-specific optimizations, so I believe it should be safe to include in OpenPBR 1.1.1.

AdrienHerubel (approved):
Thanks for the details @jstone-lucasfilm, LGTM

PR #267, opened 2025-07-28
Correct "glossy-diffuse" text in model schematic

medakk:


AdrienHerubel:
The image will be updated thoroughly in 1.2 and the change is too minor to be visible.

Issue #270, opened 2025-08-14
subsurface_color needs to be clamped to 0->1


BrianSharpe:
The OpenPBR spec states that subsurface_color should be in the range of 0 to 1, but this is not enforced by the node graph.  

This has caused an issue with the [OpenPBR Shader Playground scene](https://github.com/DigitalProductionExampleLibrary/OpenPBRShaderPlayground), where the paper surface on which the toy plane sits has a subsurface_color input in the range of ~3.0 to 4.0 (coming in through a wire).

BrianSharpe:
closing, was superceded by #276

PR #271, opened 2025-08-14
explicitly clamp subsurface_color to 0->1 range (#270)

BrianSharpe:
The OpenPBR spec says that subsurface_color should be in a strict 0->1 range,  but the current nodegraph only clamps it to being above zero (ie, allowing values above 1).  This creates a difference in look depending if a renderer allows for albedo to go above 1 or not.

This issue has come up with the (newly released) OpenPBR Shader Playground scene, where the paper surface that the airplane sits on sets the subsurface_color input to a ~3.0->4.0 range (wire input).

Better to explicitly clamp the subsurface_color to its intended 0->1 range, which will preserve look across renderers, regardless of what value is fed into it.

### DIFF
--- a/reference/open_pbr_surface.mtlx
+++ b/reference/open_pbr_surface.mtlx
@@ -410,14 +410,10 @@
       <input name="in1" type="color3" interfacename="base_color" />
       <input name="in2" type="float" interfacename="base_weight" />
     </multiply>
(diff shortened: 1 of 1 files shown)


portsmouth:
Just to note, we should also add language to the spec making it clear that this clamping has to be done by the shader. (In this case, and all other relevant cases).

Issue #272, opened 2025-08-18
emission_luminance needs to be clamped to >0


BrianSharpe:
(similar to [this](https://github.com/AcademySoftwareFoundation/OpenPBR/issues/270))

The OpenPBR spec states that emission_luminance should be in the range of >0, but this is not enforced by the node graph.

This has caused an issue with the [OpenPBR Shader Playground scene](https://github.com/DigitalProductionExampleLibrary/OpenPBRShaderPlayground), where the meetMat model has negative values coming in from a wire.

This issue is mentioned in the OpenPBR Shader Playground repro too.
https://github.com/DigitalProductionExampleLibrary/OpenPBRShaderPlayground/issues/5

BrianSharpe:
closing, was superceded by #276

PR #273, opened 2025-08-18
explicitly clamp emission_luminance to >0 range (#272)

BrianSharpe:
The OpenPBR spec states that emission_luminance should be in the range of >0, but this is not enforced by the node graph.

This has caused an issue with the [OpenPBR Shader Playground scene](https://github.com/DigitalProductionExampleLibrary/OpenPBRShaderPlayground), where the meetMat model has negative values coming in from a wire.

This commit adds an explicit >0 clamp to the incoming emission_luminance

### DIFF
--- a/reference/open_pbr_surface.mtlx
+++ b/reference/open_pbr_surface.mtlx
@@ -550,6 +550,10 @@
     </layer>
 
     <!-- Emission Layer -->
+    <max name="emission_luminance_nonnegative" type="float">
+      <input name="in1" type="float" interfacename="emission_luminance" />
+      <input name="in2" type="float" value="0.0" />
+    </max>
     <subtract name="coat_ior_minus_one" type="float">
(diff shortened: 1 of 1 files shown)


portsmouth:
Just to note, we should also add language to the spec making it clear that this clamping has to be done by the shader. (In this case, and all other relevant cases).

BrianSharpe:
> Just to note, we should also add language to the spec making it clear that this clamping has to be done by the shader. (In this case, and all other relevant cases).

I think this is already pretty clear in the spec
https://academysoftwarefoundation.github.io/OpenPBR/#parameterreference

But are you thinking something like...?
"Range: the allowed value range for scalar parameters (per-channel for color and vector parameters), which should be enforced at the shader level"

AdrienHerubel:
This PR was superseded by #276, we can work on a proposal to clarify the clamping language in a new PR if needed.

Issue #276, opened 2025-09-23
All OpenPBR inputs should get clamped/sanitized according to the parameter reference (ie section 4 of the spec)


BrianSharpe:
Issues #272 and #270 highlight the need to clamp OpenPBR inputs to what is defined in the parameter reference.  Those two issues only covered subsurface_color and emission_luminance.

After discussion on slack it was decided that it would be preferable to clamp/sanitize ALL parameters coming into OpenPBR to ensure defined and portable behavior across renderers.

Note this supersedes/replaces #272 and #270

portsmouth:
@BrianSharpe Here I assume you're referring to modifying the MaterialX reference implementation to do all the required clamps? (Rather than a modification to the spec, or do you think the spec itself needs modification as well?).

BrianSharpe:
>  assume you're referring to modifying the MaterialX reference implementation to do all the required clamps?

Yes this is right.

> Rather than a modification to the spec, or do you think the spec itself needs modification as well?

No, IMO the spec is good as it is.  It already clearly states "the allowed value range for scalar parameters" in Section 4.
https://academysoftwarefoundation.github.io/OpenPBR/#parameterreference

I've modified the MaterialX reference implementation to ensure the inputs coming through from wires (or the UI) adhere to what is stated in Section 4, which will avoid undefined behavior, ensuring better consistency and portability between renderers.

BrianSharpe:
closing, given #277 has now been merged

PR #277, opened 2025-09-23
clamp all inputs according to the parameter reference (section 4 of the spec) (#276)

BrianSharpe:
This will ensure defined and portable behavior across renderers.

Note this supersedes/replaces #270 #271 #272 and #273


AdrienHerubel (approved):
As discussed in the meetings and on Slack, there is a consensus that clamping in the nodegraph ensures consistency across implementations and discourages users to explicitly clamp inputs of the graphs.

virtualzavie:
I think the specification should reflect this decision; was it amended separately?

PR #278, opened 2025-09-30
Fix subsurface_color range to indicate 3-dimensionality

MAP233224:


AdrienHerubel (changes requested):
The change looks good, though it should target the 1.1.1 branch and needs CLA signed from contributor.

MAP233224:
> The change looks good, though it should target the 1.1.1 branch and needs CLA signed from contributor.

I really don't know what that is and can't be bothered to sign it, if you want you can delete this PR and ask someone who's already signed it to make the change properly, it's really minor.

najibna (commented):
`parametrization.md.html` now has `$[0, 1]^3$` for `subsurface_color`, which matches `color3`, but `index.html` still lists `$[0, 1]$` — those tables should stay in lockstep.

The PR also bumps spec/nodedef to v1.1, retargets `thin_film_ior`, switches `subsurface_radius_scale` to `color3`, and retunes example defaults. I'd split the type/range fix from that version-bump churn; this likely belongs on the 1.1.1 line once CLA is signed.

PR #284, opened 2025-12-08
Improved UI names for (phase function) anisotropy of transmission/subsurface

portsmouth:
Made the changes in green below. (And corresponding changes in the MaterialX file).

For the roughness anisotropies (red), we might consider "Roughness Anisotropy" instead of "Anistropy" for the UI. Though maybe in that case it's less likely to be confusing.

image


najibna (commented):
Renaming the phase-function controls to “Scatter Anisotropy” in `parametrization.md.html` and the matching MaterialX `uiname`s is a clear improvement over the old generic “Anisotropy”.

The main spec table in `index.html` still labels both as just “Anisotropy”, so those UI strings will drift unless that table is updated too.

PR #285, opened 2025-12-08
Update emission_luminance soft-max to 2000 nits.

portsmouth:
Changes as below.

Note that it seems the MaterialX file didn't have the correct `emission_luminance` default of 1000 nits, so this PR fixes that up also.

image

image


najibna (commented):
The soft-max bump to $[0, 2000]$ is applied consistently in `index.html`, `parametrization.md.html`, and `uisoftmax`. Aligning the nodedef default from `0.0` to `1000.0` matches the spec table.

Graphs that turn on `emission_weight` without setting luminance will now emit 1000 nits instead of 0 — worth a release note.

Issue #275, opened 2025-09-22
Change `transmission_scatter` to control scatter albedo directly


portsmouth:
We have [discussed ](https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1753966305282319?thread_ts=1740095087.957359&cid=C06365EAQMP)this on Slack a bit previously, but the [existing ](https://academysoftwarefoundation.github.io/OpenPBR/#model/basesubstrate/translucentbase)`transmission_scatter` parametrization is quite unintuitive. I'd like to propose we improve it for v1.2, as we still have a window to do it.

I was looking recently for example at making a volume (with grey `transmission_color` components 0.8) satisfy the white furnace test. We state in the spec that this requires `transmission_scatter` (1, 1, 1). But in fact, any (grey) `transmission_scatter` $\ge \ln$ `transmission_color` = (0.223.., 0.223.., 0.223..) works.  What happens is that the absorption is made zero when (*) $\mathbf{S} = -\ln\mathbf{C}$, and for higher $\mathbf{S}$ the absorption stays zero but the extinction changes to compensate, so the volume starts getting denser. So the density of the volume is dependent on the scattering color.

(*) with $\mathbf{C}$=`transmission_color` and $\mathbf{S}$=`transmission_scatter`.

I show below the behavior of the current volume compared to dense SSS, with:

   - `transmission_depth` 0.11157, `transmission_color` (0.8, 0.8, 0.8) (so extinction is 2 for $\mathbf{S} \le 0.223$). This is a sort of typical intermediate density volume, for a thick liquid etc.
   - `subsurface_radius` 0.01 (so extinction is 100), so very dense volume as SSS is designed for.
 
for various "scatter color" (i.e. `transmission_scatter` for volume or `subsurface_color` for SSS) of the form (R, 0.5, 0), plus white furnace tests.

For comparison, in the "volume (proposed)" column I show the result of making the `transmission_scatter` color define the (single-scattering) albedo of the volume directly:

|  scatter color  |    dense SSS    |     volume (current)  |     volume (proposed) |
|:------------:|:-------------:|:-------------:|:-------------:|
[...]

PR #286, opened 2025-12-08
`transmission_scatter` sets volume albedo directly

portsmouth:
Changes as below.  See https://github.com/AcademySoftwareFoundation/OpenPBR/issues/275 for the difference in appearance, which is clearly more intuitive.

This gets rid of the previous rather clumsy adjustment of the absorption and extinction to accommodate the arbitrarily high scattering coefficient. The MaterialX graph is correspondingly simplified.

As discussed in https://github.com/AcademySoftwareFoundation/OpenPBR/issues/275, this is now also more consistent with the existing parametrization since the `transmission_scatter` color is an albedo so the $[0,1]$ range of the components is correct (whereas before the $[0,1]$ range didn't make sense).

image


portsmouth:
It would be helpful if the reviewers can verify that the changes to the MaterialX graph are correct.

AdrienHerubel:
As discussed in the meeting, we are now considering this look breaking change for 1.2. @peterkutz can you please formally review ?

fpsunflower (approved):
Makes sense to me.

peterkutz (approved):
Reviewed and approved! I've been looking forward to this simplification. Thanks for officially updating the spec @portsmouth .

I have not carefully reviewed the .mtlx changes because I'm not particularly familiar with MaterialX syntax.

AdrienHerubel (approved):
There is a consensus that this PR should be merged and has been reviewed and approved.

Issue #274, opened 2025-09-09
State orthonormalization policy for input normals and tangents


portsmouth:
This issue was highlighted by Arnon Marcus in a Slack [discussion](https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1757198799809719):

>I have a question about tangent vs normal inputs. For each pair (coat or geometry), is it expected that they are always provided already perpendicular to each over? If not,  is the normal expected to be the "primary" one and the tangent is expected to be realigned to the normal internally by the material? What is the exact expectation there? 

In my view, I agree that we should really explain in the specification how the supplied shading normal N and anisotropy tangent T are supposed to be made consistent, if they are not orthogonal.

In detail, I would propose we assume in general the shader is supplied

- The  final bumped/perturbed shading normal $N'$.
- The desired tangent vector defining the specular reflection anisotropy, say $T'$.
- (These will usually be based on some reference normal $N$ and reference tangent $T$, via normal and flow mapping, but the shader does not necessarily have direct access to those).

Then we state that we do not assume that the supplied $N'$ and  $T'$ are necessarily orthonormal, but if they are not, the shader should make $T'$  orthonormal to $N'$, by projecting 
```math
T' \rightarrow T' - (N'\cdot T') N'
```
and renormalizing.

Then the proejcted $T'$ used for anisotropy lies in the local $xy$-plane, which makes sense for the microfacet theory we discuss.

portsmouth:
One point to make though is that the proposed projection of $T'$ onto the $N'$ plane alters the tangent vector field away from original $T'$, which presumably was designed to lie in the plane normal to the $N$ before bump mapping was applied (giving the flow map *before* bumping the geometry).  

Of course if $T'$ was carefully authored to follow the bumped surface not the original geometry, it would already be orthogonal to $N'$. (Maybe tools that allow users to paint flow maps via vector fields do account for bump, but I guess the bump map could always get changed).

It's not necessarily obvious that the direct projection onto the $N'$ plane is the best choice of $T'$ direction in the plane (it seems a reasonable interpretation though). It at least preserves the continuity of curves defined relative to the $T$ frame, if not the derivative.

PR #287, opened 2025-12-09
Clarify orthonormalization policy for normals and tangents

portsmouth:
To address https://github.com/AcademySoftwareFoundation/OpenPBR/issues/274, I updated the language in the spec to be more explicit that the input normals and tangents are expected to be made orthonormal (if it is not known that they already are):

image

<BR>
<BR>

For a reference to the concept of "2D flow maps", I cited the paper "[Water Flow in Portal 2](https://advances.realtimerendering.com/s2010/Vlachos-Waterflow(SIGGRAPH%202010%20Advanced%20RealTime%20Rendering%20Course).pdf)" by Vlachos et. al (2010), as this seems to have the nicest discussion I could find of this.

It also seems to make sense to connect the microfacet discussion with the normal and tangent inputs more explicitly (e.g. the normal inputs are controlling the *macronormal* [*]):

image

<BR>  

image

[*] Actually, technically the "mesonormal" in Heitz' nomenclature (meaning the center of the micronormal distribution, which does not necessarily coincide with the geometric macronormal, e.g. in LEAN/LEADR mapping), but it seems too technical/academic to use this term here.


index.html
```diff
@@ -1248,20 +1251,25 @@
...
 
-The perturbed normals and tangents will of course be specified by input textures (or possibly procedurally). As noted in the Metadata section, the particular parametrization which maps the contents of the texture to the perturbation of the normal or tangent in the shading frame is not defined in the model itself, as this can be done in multiple ways. It is instead just assumed that the material is packaged with metadata that makes this mapping unambiguous.
+It is _not_ required that the resultant input **`geometry_normal`** $\mathbf{N}_B$ and **`geometry_tangent`** $\mathbf{T}_B$ are strictly normalized and orthogonal (and similarly for the coat normal and tangent).
+For example reconstruction filtering techniques such as texture filtering may cause the interpolated normal to not be normalized. However, we _do_ require that the implementation should (if not already ensured) ensure orthonormality of the final vertex shading frame of the BSDF, by applying
```
portsmouth: Just to note, this is equivalent to [Gram-Schmidt ](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process) orthogonalization. We might as well state that.

Done in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/287/commits/15bf2e1c95989c2ec5991c3ae2643a1e21857c6b

najibna (commented):
The Gram-Schmidt policy is the useful part: inputs need not already be orthonormal, but the BSDF shading frame must be.

Nits: “pertubed” should be “perturbed”, and $\mathbf{T}_B \rightarrow \mathrm{normalize}(\mathbf{T}_B - (\mathbf{N}_B\cdot\mathbf{T}_B)\mathbf{N}_B)$ is only correct once $\mathbf{N}_B$ is unit length. Please state the order (normalize N, then project T) and what to do if T is parallel to N.

Issue #269, opened 2025-08-13
TODOs following SIGGRAPH write-ups


portsmouth:
Various problems with the current spec have been noticed and discussed as a result of the work on the write-ups for SIGGRAPH. I'll collect those here for reference, and PRs can be created as needed following further discussion.

- **Coat darkening (implementation)**. I noticed an error in the coat darkening code (in the numerator of the darkening factor), which I fixed in the pseudo-code listing in the write-up. This will have affected the look particularly with colored base and coat. This seems pretty important to get right, since the darkening is on by default, and we don't want it producing artifacts. We need to:
     - check the new darkening code theory and renders 
     - implement the correction in MaterialX (and our internal implementations)
     - fix up the recommended formulas in the spec (or just omit them and refer to the new document).
  
- **Coat darkening (specification)**. There is an inconsistency in the current (v1.2) proposed statement of how the coat darkening parameter works, since we define it in terms of an albedo which is only sensible in the perfectly smooth case. This can be generalized with better wording though, which will also bring it in line with the more careful statements made for subsurface and glossy-diffuse.

- **Coat darkening (theory)**. The formula we state for the darkening effect in the case of a smooth base is probably incorrect. In this limit there is an exact solution (for a smooth coat), which we should properly work out. 

- **Thin-walled mode**.  It was noted that the spec is currently not clear enough for an implementor in various aspects. See Slack threads [here ](https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1749690787065219)and [here](https://academysoftwarefdn.slack.com/archives/C06365EAQMP/p1755051786659599?thread_ts=1754967652.973689&cid=C06365EAQMP). In particular:
   - generally the implementation is not super obvious (of the ladder of reflections in the absorbing, rough thin-wall), since we don't reference a standard paper (there isn't one). A write-up of a recommended model would be helpful.
   - In the subsurface case, it's not stated correctly since the dielectric/spec Fresnel reflection has to still happen, but we don't make it clear how. Really we need to say the thin-wall is embedded (or bounded) by surrounding dielectric interfaces, more explicitly. (This likely won't impact implementations though, since they are already implicitly made that assumption). 
   - `transmission_depth` is ignored in this mode, which is not clear.
[...]

PR #288, opened 2025-12-09
Clarified wording following SIGGRAPH report

portsmouth:
This attempts to address the outstanding points discussed in https://github.com/AcademySoftwareFoundation/OpenPBR/issues/269.

These were already addressed in https://github.com/AcademySoftwareFoundation/OpenPBR/pull/253:

  - Coat darkening (implementation). The formula was fixed.
  - Coat darkening (specification). Wording was suitably generalized.
 -  Coat darkening (theory). Smooth base case was worked out and correctly implemented.

Here I address the other issues.

## Thin-walled mode. 

It was a bit unclear in the subsurface case that there is intended to be a dielectric slab *embedding* the thin-wall SSS "dense sheet", where the boundaries of the dielectric provide the specular lobe as usual.

The new text makes that more explicit. An issue is then that technically, light should bounce around inside the slab between the dense sheet and the dielectric walls, causing darkening. We could attempt to define something to get rid of that, but it gets too technical I think. I feel the best approach here is to simply say that that those effects should in theory be accounted for. 

We already said the same effectively for the translucent dielectric case, since we said:

>The translucent-base slab can be considered an infinitesimally thin sheet of dielectric (absorbing but non-scattering), with the BSDF fdielectric on both sides. *A ladder of inter-reflections occurs inside this slab producing a reflected lobe and un-deflected refracted lobe*.

That ladder generates IOR-dependent darkening. It seems consistent to simply mention that this similar physical effect occurs in the subsurface case, without adding any special mechanism to get rid of it. In practice, implementations could reasonably just ignore this (e.g. MaterialX), though this produces some inaccuracy/unphysicality in the appearance. We could perhaps work out a nicer approximation to apply, and provide it in the MaterialX graph though. 

I also made it more explicit that the diffuse lobes of the subsurface model are _required_ to be EON diffuse lobes, which seems reasonable (and already the case in the MaterialX graph).

image

--------------------------------------------------------

As I noted in the technical report, I really feel that this thin-wall model is too loosely defined still at present. A problem is that there is no one publication that works out the necessary BSDFs in full detail. It also seems a bit ad-hoc to have these two separate models of "dielectric sheet with embedded stufff", with either no scattering or dense scattering (the latter a rather crude approximation as well, with Oren-Nayar lobes involved).

It would be better to have available a more general "thin-slab of arbitrary volume" as a primitive:
>Currently we treat the thin-wall limit of the translucent volume
and the subsurface as two distinct cases (essentially absorbing only and scattering-dominated, respectively).
However, it would be useful to have a more general model where the thin-wall interior medium is a scattering
volume with configurable transmittance and albedo. If the scattering is not totally dominant then there is an intermediate
regime between “thin-wall dielectric” (which roughens due to the refractive boundaries but does not
scatter), and “thin-wall subsurface” that produces essentially diffuse reflection and transmission lobes. However,
at present, there isn’t a well-defined model for this intermediate case. Furthermore, this generalized thin-wall
model would make more sense in the context of a unified base volume, which is also considered future work.

This will have to be left for future work though. At present, this is as close as we could get to a completely physically specified thin-wall model.

## Subsurface and glossy-diffuse albedos

> In the discussion of the meaning of base_color for the glossy-diffuse slab, and subsurface_color, we talk about them controlling the "multi-scatter" albedo, but we don't make clear how the surrounding media (coat etc.) are involved. Probably we need to say that these definitions (of the meaning of the colors) are done in the vacuum exterior case, as e.g. we don't want the underlying albedos of the base adjusting as the coat properties change.

I think a reasonable fix for these is simply to say explicitly that the definitions made are supposed to be applied assuming vacuum exterior (for the subsurface or glossy-diffuse slabs). This serves to define the BSDFs and media, then the more general case involving superimposed layers will follow from that.

image

--------------------

image

<BR>
<BR>
<BR>
<BR>

(These definitions still seem extremely academic, and I'm skeptical that anyone has actually studied them and modified their implementation accordingly. I'm assuming almost everyone, or actually everyone, simply uses an approximate albedo-scaling approach. The current text is an attempt to make it more physically general than simply assuming albedo-scaling based layering approximation, but done in a way which ensures that albedo-scaling is not a bad approximation. Perhaps there is a better way to define this, at a later stage, that seems less esoteric).


najibna (commented):
Calling out a bare slab with vacuum exterior for `base_color` and `subsurface_color` is the right way to keep those albedos independent of coat/fuzz.

The thin-wall subsurface rewrite (dense sheet embedded in a finite-thickness dielectric, EON via `base_diffuse_roughness`) is clearer, but it does tighten the spec for renderers still on classic Oren-Nayar. The new “should” darken/saturate from inter-reflections has no formula — better marked recommended than implied-normative.

PR #293, opened 2026-01-30
Fix typo in reference MaterialX graph

jstone-lucasfilm:
This changelist fixes a typo in the reference MaterialX graph for OpenPBR surface, where the return value of a BSDF mix needs to be of BSDF type.


portsmouth:
Was this causing the graph to fail to compile?

jstone-lucasfilm:
Yes, exactly.  This prevents the graph from being used in rendering or shader translation, as MaterialX is strongly typed.

AdrienHerubel (approved):
LGTM

PR #295, opened 2026-03-03
Fix CHANGELOG.md

portsmouth:
New changelog renders as below:

image


najibna (commented):
Much easier to scan with titles and PR links. Two small consistency notes:
- `#231` (`emission_weight`) looks like a new parameter, so ✨ fits better than 🎨.
- The `#277` bullet still leads with leftover “This will ensure defined and portable behavior” instead of the actual clamp.

PR #300, opened 2026-04-15
Fix 1.1.1: metal edge tint weight and thin-walled geometry flag

portsmouth:
Two bug fixes for the 1.1.1 point release:

- Metal edge tint incorrectly multiplied by specular_weight (cherry-picked from AcademySoftwareFoundation/OpenPBR#240): the color82 (edge tint) input on metal_bsdf and metal_bsdf_tf was routed through a multiply node that applied specular_weight. Since the whole lobe is already weighted by specular_weight, the edge tint was being double-weighted. Now connects directly to the specular_color interface input.

- Connect geometry_thin_walled flag to <surface> (borrowed from AcademySoftwareFoundation/MaterialX#2759): the thin-walled geometry flag was not being passed through to the surface node in the MaterialX graph.


jstone-lucasfilm (approved):

PR #301, opened 2026-04-15
Add MaterialX validation script and CI workflow

portsmouth:
## Summary

- Adds `validation/validate_mtlx.py`, a script that validates `.mtlx` files against the MaterialX standard libraries
- Adds `.github/workflows/validate-mtlx.yml`, which runs the validator on `reference/open_pbr_surface.mtlx` and all `examples/*.mtlx` on every PR or push that touches `.mtlx` files

## How it works

The validator uses `pip install materialx` (no build-from-source required), loads the standard libraries, and runs `doc.validate()`. It also performs an explicit nodename reference check on the standalone document — this catches dangling connections that the with-libraries validation misses when a file shares a name with a standard library file (as `open_pbr_surface.mtlx` does with `libraries/bxdf/open_pbr_surface.mtlx`).

## Test plan

- [ ] Workflow triggers on PRs that modify `.mtlx` files
- [ ] `reference/open_pbr_surface.mtlx` passes
- [ ] All `examples/*.mtlx` pass
- [ ] A PR that introduces a broken nodename reference fails CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)


najibna (commented):
The explicit nodename check plus `check_library_shadowing()` is the part that actually matters — `reference/open_pbr_surface.mtlx` shares a basename with the MaterialX stdlib copy, so `doc.validate()` alone can miss dangling connections.

`validation/validate_mtlx.py` is a large general linter, and `pip install materialx` is unpinned. `--version` is also a `store_true` next to a required positional `file`, so `python validation/validate_mtlx.py --version` fails. A thin wrapper around `doc.validate()` plus the shadowing check would be easier to maintain.

PR #303, opened 2026-04-17
Merge v1.1.1 development to main

jstone-lucasfilm:
This changelist merges v1.1.1 development from dev_1.1.1 to main, in preparation for marking the release of OpenPBR v1.1.1.


AdrienHerubel (approved):
I see at least one commit in the merge PR belonging to the dev_1.1  (#217) which should have been in main already, so it is possible that main was a bit too much behind.

PR #309, opened 2026-07-10
Fix missing link to some sections

RiverIntheSky:
The spec mentions "dielectric base" section in multiple places, but there was no section called "dielectric base"


RiverIntheSky:
**Before:**
Screenshot 2026-07-10 at 15 13 48
**After:**
Screenshot 2026-07-10 at 15 13 53

najibna (commented):
The new `### Dielectric base` heading is the right fix — Markdeep can only link that name if the heading exists (`Translucent base` already had one).

Nit: `###  Dielectric base` has a double space after `###`, unlike the other headings. EasyCLA is also still unsigned on this PR.

PR #313, opened 2026-07-14
Fix: invalid OpenPBR parameterizations in some mtlx examples

sherholz:
Some of the `.mtlx` example files contain color values > 1.0. 
Since it is unclear how different renderer handle these out of range values
the rendering outcome is unclear/ill-defined.
This PR normalizes these color values so that the max value is 1.0.


najibna (commented):
Normalizing `specular_color` by the max channel is the right way to stay in `[0,1]^3` without changing hue. Splitting milk/petroleum into `subsurface_radius` plus a unit-normalized `subsurface_radius_scale` also matches the spec.

Two small follow-ups:
- Trailing spaces in `open_pbr_potassium.mtlx` (`0.977 ,`) and `open_pbr_rubidium.mtlx` (`0.938 ,`).
- `emission_weight=1` is added only on the 2700K bulb. If the 1.2 default is 0, the 5000K and LCD emitters would go dark and need the same input.

examples/open_pbr_light_bulb_2700k.mtlx
```diff
@@ -4,6 +4,7 @@
     <input name="surfaceshader" type="surfaceshader" nodename="open_pbr_surface_surfaceshader" />
   </surfacematerial>
   <open_pbr_surface name="open_pbr_surface_surfaceshader" type="surfaceshader">
+    <input name="emission_weight" type="float" value="1.0" />
```
AdrienHerubel:
```suggestion
```
Removing the emission_weight parameter as it is a 1.2 parameter.

AdrienHerubel (approved):
As discussed at the meeting, the changes make sense. I removed one of the 1.2 parameters that was added to an example, it could be added back in the 1.2 branch.

PR #315, opened 2026-09-21
Capitalize equation references

### LINKED ISSUE #263: Capitalize all equation names
Some equation references in the spec are lowercase while some are uppercase. To be consistent with figure references and technical paper standards, and to clearly show that these references are names of particular formulas, I propose capitalizing all equation names. 

For example, "equation 18" would become "Equation 18".

najibna:
Fixes #263

Capitalizes equation references throughout the OpenPBR specification for
consistency with figure references and technical writing conventions.

No functional changes.


AdrienHerubel (approved):
The proposed changes make the specification more consistent, of the 38 occurrences of the word Equation, only 6 started with an upper case.

